PR #9, opened 2021-11-08
Modify Via git clone (dev install)

csh519:
Add comment for users who only want to use basic features not Ml validation feature.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,8 @@ $ pip install git+https://github.com/Samsung/CredSweeper.git
 ``` bash
 $ git clone https://github.com/Samsung/CredSweeper.git
 $ cd CredSweeper
-$ pip install -qr requirements.txt
+# Annotate "tensorflow" if you don't want to use the ML validation function.
+$ pip install -qr requirements.txt 
 ```
 
 ### Run
--- a/README.md
+++ b/README.md
@@ -32,10 +32,8 @@ CredSweeper is a tool to detect credentials in any directories or files. CredSwe
 
 Stable (release 1.0.0):
 ``` bash
-# Without Ml validation feature
-$ pip install git+https://github.com/Samsung/CredSweeper.git@v1.0.0
 # With Ml validation feature
-$ pip install CredSweeper[ml]@git+https://github.com/Samsung/CredSweeper.git@v1.0.0
+$ pip install git+https://github.com/Samsung/CredSweeper.git@v1.0.0
 ```
 
 Latest (current master branch):
--- a/README.md
+++ b/README.md
@@ -32,12 +32,18 @@ CredSweeper is a tool to detect credentials in any directories or files. CredSwe
 
 Stable (release 1.0.0):
 ``` bash
+# Without Ml validation feature
 $ pip install git+https://github.com/Samsung/CredSweeper.git@v1.0.0
+# With Ml validation feature
+$ pip install CredSweeper[ml]@git+https://github.com/Samsung/CredSweeper.git@v1.0.0
 ```
 
 Latest (current master branch):
 ``` bash
+# Without Ml validation feature
 $ pip install git+https://github.com/Samsung/CredSweeper.git
+# With Ml validation feature
+$ pip install CredSweeper[ml]@git+https://github.com/Samsung/CredSweeper.git
 ```
 
 #### Via git clone (dev install)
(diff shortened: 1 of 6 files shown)


meanrin (commented):
Removing `tensorflow` will result in error at import time

```
  File "CredSweeper/credsweeper/ml_model/ml_validator.py", line 9, in <module>
    from tensorflow.keras import models
ModuleNotFoundError: No module named 'tensorflow.keras'
```

That's due to the fact that `ml_validator.py` and `tensorflow` imported unconditionally

We need to either:
a) Move `from credsweeper.ml_model import MlValidator` under `if ml_validation:` in `app.py`
b) Create `try: ... except ImportError: ` around `tensorflow` imports in `ml_validator.py`

Please also note that `numpy` `scipy` `sklearn` is also not required if ML is not used

csh519:
> That's due to the fact that `ml_validator.py` and `tensorflow` imported unconditionally
> 
> We need to either: a) Move `from credsweeper.ml_model import MlValidator` under `if ml_validation:` in `app.py` b) Create `try: ... except ImportError: ` around `tensorflow` imports in `ml_validator.py`
> 
> Please also note that `numpy` `scipy` `sklearn` is also not required if ML is not used

@meanrin 

Thank you for comment!
I moved `MlValidator` import to under `if ml_validation:` 40ecf30.
And added `numpy`, `scipy`, and `scikit-learn` on annotation list 9aef185.

meanrin (approved):
LGTM

meanrin (commented):
We can add similar thing with pip install via `extras_require` option

https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#optional-dependencies

For example we can have:
```$ pip install git+https://github.com/Samsung/CredSweeper.git```
for default "no-ml" install

and

```$ pip install git+https://github.com/Samsung/CredSweeper.git[ml]```
For ml

**Comment by csh519** [comment]
> We can add similar thing with pip install via `extras_require` option
> 
> https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#optional-dependencies
> 
> For example we can have: `$ pip install git+https://github.com/Samsung/CredSweeper.git` for default "no-ml" install
> 
> and
> 
> `$ pip install git+https://github.com/Samsung/CredSweeper.git[ml]` For ml

@meanrin 

Thank you for suggestion.
I added few commits regarding `extras_require` option.
Please review. 😃

**Review by meanrin — CHANGES_REQUESTED** [review]
Please add  `from credsweeper.ml_model import MlValidator` in `app.py` line 152, after `if self.config.ml_validation`, otherwise it wont work :(

Other than that LGTM, works really well!

**Comment by meanrin** [comment]
We would need to include change to pip in next release patch notes, otherwise it might hurt people who already using ML validation feature via pip install

File: README.md [hunk 1]
```diff
@@ -32,20 +32,27 @@ CredSweeper is a tool to detect credentials in any directories or files. CredSwe
 
 Stable (release 1.0.0):
 ``` bash
+# Without Ml validation feature
 $ pip install git+https://github.com/Samsung/CredSweeper.git@v1.0.0
+# With Ml validation feature
+$ pip install CredSweeper[ml]@git+https://github.com/Samsung/CredSweeper.git@v1.0.0
```
meanrin, line 38: Hm, I think this line should be removed
v1.0.0 setup file do not contain separate Ml extra:

https://github.com/Samsung/CredSweeper/blob/v1.0.0/setup.py

csh519:
> Please add `from credsweeper.ml_model import MlValidator` in `app.py` line 152, after `if self.config.ml_validation`, otherwise it wont work :(
> 
> Other than that LGTM, works really well!

@meanrin 

Actually I moved import line by 40ecf30, please check. I think it is working well.

README.md
csh519: Thank you for review.
I removed the line by afb2d45, please check.

meanrin:
@csh519 
Current code would produce the next error:
```
  File "CredSweeper/credsweeper/app.py", line 159, in post_processing
    if MlValidator.validate_group(value, group_candidates):
NameError: name 'MlValidator' is not defined
```

Python interpret imported objects as variables/pointers, so even if we import `MlValidator` in `__init__` function it would not be visible in any other function, as garbage collector would kill this pointer as soon as we step out of `__init__` local namespace
`MlValidator` as a module would still be present in memory and properly imported, it's just variable with exact name `MlValidator` declared on line 60 would die

meanrin:
@csh519 
Btw, we don't really need `MlValidator` module anywhere but `post_processing` function, so probably we can just delete all `if ml_validation: ...` from `__init__` func

csh519:
> @csh519 Btw, we don't really need `MlValidator` module anywhere but `post_processing` function, so probably we can just delete all `if ml_validation: ...` from `__init__` func

@meanrin 

I agree. I moved the code block by 634bf21, please check.
And there were TC failure cases.. so I fixed with 7fb415d.

meanrin (approved):
LGTM!

Also, remember bug with tensorflow 2.6?
Bug is still there, but they released 2.7 few days ago. And 2.7 have this bug fixed
I'll update setup with future PR so 2.7 would be allowed as dependency, but 2.6 still not

Dmitriy-NK (approved):
Thanks for updates. LGTM

PR #16, opened 2021-11-15
Add test action

meanrin:
Add git-hub action for testing cred sweeper

This work will test:
```shell
pip install .
pip install .[ml]
pytest -m pytest -m "not api_validation" tests/
```

On Python: 3.7, 3.8, 3.9
Latest stable Windows, MacOS, Ubuntu

API validation tests omitted so we do not DDOS services from GitHub servers

PR #17, opened 2021-11-15
List owners by account in CODEOWNERS

meanrin:
Current opting with using group name as code owners is nice and compact
Unfortunately it do not list group users in "Reviewers" so even if you notified: you still need to request this review on yourself, before you can lest a review 

image

This PR change this back

It's optional quality of life improvement

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,1 @@
-* @Samsung/CredSweeper_Maintainers
+* @silentearth @csh519 @yuzzyuzz @meanrin @Dmitriy-NK @ARKAD97


csh519:
@meanrin 

> Unfortunately it do not list group users in "Reviewers" so even if you notified: you still need to request this review on yourself, before you can lest a review

Could you explain it more? I can't understand what you mean actually.

meanrin:
> Could you explain it more? I can't understand what you mean actually.

Button for leaving review would not appear on the main ("Conversation") page unless you added to review by the name 
But button still will be present on the "File changes" page even if you not added, so maybe not such a big deal

And it's just think it's weird how reviews behave with this groups...
For example after you left a comment in #16, group disappeared from "Reviewers" list. So I added some people back manually 
That didn't happen in this PR for some reason, so idk...

csh519 (approved):
@meanrin 

> Button for leaving review would not appear on the main ("Conversation") page unless you added to review by the name But button still will be present on the "File changes" page even if you not added, so maybe not such a big deal
> 
> And it's just think it's weird how reviews behave with this groups... For example after you left a comment in #16, group disappeared from "Reviewers" list. So I added some people back manually That didn't happen in this PR for some reason, so idk...

Hmm.. I agree with you. It would be better to roll back as before.

LGTM :+1:

meanrin:
It seems i was wrong, or it was rust my local bugs...
We tested with @Dmitriy-NK , and it seems that `you still need to request this review on yourself, before you can lest a review` was some kind of my local bug and cannot be reproduced
And all people from group list still receive email notifications, so it's all that matters

PR #22, opened 2021-11-24
Fix diff Out of Bound error

ARKAD97:


meanrin (approved):
This fixes out of bound issue with diff scan

Dmitriy-NK (approved):
Thanks for quick fix

PR #27, opened 2021-11-26
Update patch anlisys for support multiline rules

Dmitriy-NK:
Update patch anlisys for support multiline rules. 
Previously row indexes for second line in multiline rule was incorrect.
 
Also, added test case for multiline rules handling and content providers

### DIFF
--- a/credsweeper/utils/util.py
+++ b/credsweeper/utils/util.py
@@ -109,7 +109,7 @@ class Util:
         return {}
 
     @classmethod
-    def preprocess_file_diff(cls, changes: List[Dict], change_type: str) -> Tuple[List[int], List[str]]:
+    def preprocess_file_diff(cls, changes: List[Dict]) -> List[Tuple[str, int, str]]:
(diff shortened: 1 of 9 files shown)


credsweeper/utils/util.py
```diff
@@ -1,12 +1,29 @@
...
+class DiffRowData:
+    """Class for keeping data of diff row."""
+    line_type: str
+    line_numb: int
+    line: str
+
+    def get_data(self) -> float:
+        return self.unit_price * self.quantity_on_hand
+
+    # def __init__(self, line_type: str, line_numb: int, line: str):
```
meanrin, line 23: Please remove commented code

meanrin (approved):
This PR fixes few issues with multi-line credentials in diff scan
Namely:
- fix that sometimes wrong line number returned 
- add much more diff tests

csh519 (approved):
LGTM 👍

Issue #26, opened 2021-11-26
Regex for pypi tooken


meanrin:
Current pypi token in our rules in overly specific:
https://github.com/Samsung/CredSweeper/blob/01b5f2abb5e38a77d00c914163ca72f15b228ea7/credsweeper/rules/config.yaml#L338

But here is a real token example, and it starts from

```
pypi-AgENdGVzdC5weXB
```

So it cannot be matched.

I assume it might be better to just change regex to 
```
 - (?P<value>pypi-[\w_\-]{167})
```

PR #28, opened 2021-11-26
Update pypi regex

meanrin:
Related to https://github.com/Samsung/CredSweeper/issues/26

Update pypi regex, as current regex cannot detect some real tokens

- Old regex included `AgEIcHlwaS5vcmc` at the start, which is not always the case for real tokens
- Real tokens can have different lengths

### DIFF
--- a/credsweeper/rules/config.yaml
+++ b/credsweeper/rules/config.yaml
@@ -335,7 +335,7 @@
   severity: high
   type: pattern
   values:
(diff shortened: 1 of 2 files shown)


csh519 (approved):
Thank you for updating pypi regex.

LGTM 👍

Dmitriy-NK (approved):
LGTM

ARKAD97 (approved):
My mistake, thanks for fix.

PR #34, opened 2021-12-01
Add model train code

meanrin:
- Add code to train ML model
- Add readme on how to run it
- Add code to augment cred data with realistic passwords


experiment/src/prepare_data.py
```diff
@@ -0,0 +1,34 @@
...
+import subprocess
+
+
+def execute_scanner(dataset_location: str, result_location_str, j, use_ml=False):
+    """Execute CredSweeper as a separate process to make sure no global states is shared with training script"""
+    dir_path = os.path.dirname(os.path.realpath(__file__)) + "/.."
+    command = f"{sys.executable} -m credsweeper --path {dataset_location}/data --save-json {result_location_str} -j {j}"
+    if use_ml:
+        command += " --ml_validation"
+    subprocess.call(command, shell=True, cwd=dir_path, stdout=subprocess.DEVNULL, stderr=subprocess.STDOUT)
```
ARKAD97, line 12: Why not to use credsweeper as a library?

meanrin: Oh, it's to guaranty total independence of memery
CredSweaper still have `MlValidator` as a global (static) object, and we use TF/Keras. TF/Keras also like to spam global objects. So I'm scared that it will somehow influence the training results

csh519 (approved):
Thank you for adding `experiment` module.

LGTM :+1:

Dmitriy-NK (approved):
Looks good

PR #29, opened 2021-11-30
Update docs examples in the Readme and add contribution guideline

Dmitriy-NK:
Fixes #18 

Add contribution document with Exact step-by-step description on how to create a new PR 
Add link to MIT license official site in places where we mention it
Edit install/run/tests examples in the Readme, such that code can be copied and executed without any change
(currently we have some lines that start from # or $ that should be manually removed)

PR #37, opened 2021-12-03
Fix urls due to doc -> docs rename

meanrin:
Fix urls due to doc -> docs rename with #29


Dmitriy-NK (approved):
Thank you for fixes

ARKAD97 (approved):
Thanks

PR #36, opened 2021-12-01
Add UTF-16 encoding support for patch files

Dmitriy-NK:
Fix unicode decode error that can happen with diff scan 
```
Traceback (most recent call last):
File "/usr/lib/python3.8/runpy.py", line 194, in _run_module_as_main
     return _run_code(code, main_globals, None,
   File "/usr/lib/python3.8/runpy.py", line 87, in _run_code
     exec(code, run_globals)
   File "/home/avas/.local/lib/python3.8/site-packages/credsweeper/__main__.py", line 107, in <module> main()
   File "/home/avas/.local/lib/python3.8/site-packages/credsweeper/__main__.py", line 99, in main scan(args, content_provider, added_json_filename)
   File "/home/avas/.local/lib/python3.8/site-packages/credsweeper/__main__.py", line 82, in scan credsweeper.run(content_provider=content_provider)
   File "/home/avas/.local/lib/python3.8/site-packages/credsweeper/app.py", line 92, in run
     file_extractors = content_provider.get_scannable_files(self.config)
   File "/home/avas/.local/lib/python3.8/site-packages/credsweeper/file_handler/patch_provider.py", line 57, in get_scannable_files
     diff_data = self.load_patch_data()
   File "/home/avas/.local/lib/python3.8/site-packages/credsweeper/file_handler/patch_provider.py", line 37, in load_patch_data
     raw_patches.append(Util.read_patch_file(file_path))
   File "/home/avas/.local/lib/pytho
   File "/usr/lib/python3.8/codecs.py", line 322, in decode
     (result, consumed) = self._buffer_decode(data, self.errors, final)
 UnicodeDecodeError: 'utf-8' codec can't decode byte 0xf6 in position 983: invalid start byte
```
Add support of UTF-16 encoding
Add more meaningful logging message that file can't open
Add corresponding test cases

PR #38, opened 2021-12-06
Update read file content function

Dmitriy-NK:
Update file content read. Combine patch file reading and full-text file reading

### DIFF
--- a/credsweeper/file_handler/text_content_provider.py
+++ b/credsweeper/file_handler/text_content_provider.py
@@ -1,8 +1,9 @@
+from os import read
 from typing import Dict, List, Optional
 
 from credsweeper.file_handler.analysis_target import AnalysisTarget
 from credsweeper.file_handler.content_provider import ContentProvider
-
+from credsweeper.utils import Util
 
 class TextContentProvider(ContentProvider):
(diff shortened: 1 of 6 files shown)


credsweeper/file_handler/text_content_provider.py
```diff
@@ -1,8 +1,9 @@
+from os import read
 from typing import Dict, List, Optional
 
 from credsweeper.file_handler.analysis_target import AnalysisTarget
 from credsweeper.file_handler.content_provider import ContentProvider
-
+from credsweeper.utils import Util
 
 class TextContentProvider(ContentProvider):
```
meanrin, line 8: I think there are should be 2 empty lines between class and import, please check

`PEP 8: E302 expected 2 blank lines, found 1`

tests/file_handler/test_text_content_provider.py
```diff
@@ -1,4 +1,5 @@
 from pathlib import Path
+from unittest import mock
```
meanrin, line 2: `mock` is not used in this file

ARKAD97 (approved):
LGTM!

meanrin (approved):
This PR is similar to the #36 but for non-diff files
LGTM

Issue #43, opened 2021-12-07
Too much info from `WARNING | util | UnicodeError`


meanrin:
Running scanner on repo with binary files can easily transform CLI out to

```python
2021-12-07 16:17:11,870 | WARNING | util | UnicodeError: Can't read content from "/mnt/data/datasets/validate_repos/node/test/fixtures/wpt/encoding/legacy-mb-tchinese/big5/big5_errors.html" as utf16.
2021-12-07 16:17:11,903 | WARNING | util | UnicodeError: Can't read content from "/mnt/data/datasets/validate_repos/node/test/fixtures/wpt/encoding/legacy-mb-tchinese/big5/big5_chars-csbig5.html" as utf8.
2021-12-07 16:17:11,903 | WARNING | util | UnicodeError: Can't read content from "/mnt/data/datasets/validate_repos/node/test/fixtures/wpt/encoding/legacy-mb-tchinese/big5/big5_chars-csbig5.html" as utf16.
2021-12-07 16:17:11,918 | WARNING | util | UnicodeError: Can't read content from "/mnt/data/datasets/validate_repos/node/test/wasi/wasm/create_symlink.wasm" as utf8.
2021-12-07 16:17:11,930 | WARNING | util | UnicodeError: Can't read content from "/mnt/data/datasets/validate_repos/node/test/wasi/wasm/create_symlink.wasm" as utf16.
2021-12-07 16:17:11,960 | WARNING | util | UnicodeError: Can't read content from "/mnt/data/datasets/validate_repos/node/test/fixtures/wpt/encoding/resources/two-boms-utf-16be.html" as utf8.
2021-12-07 16:17:11,960 | WARNING | util | UnicodeError: Can't read content from "/mnt/data/datasets/validate_repos/node/test/fixtures/wpt/encoding/resources/two-boms-utf-16le.html" as utf8.
2021-12-07 16:17:11,964 | WARNING | util | UnicodeError: Can't read content from "/mnt/data/datasets/validate_repos/node/test/fixtures/wpt/encoding/resources/utf-32-big-endian-bom.html" as utf8.
```

This is not as useful, especially due to the fact that utf-16 files would always create this log message even thou they would be read correctly 

Propose to change this log level to INFO or DEBUG
https://github.com/Samsung/CredSweeper/blob/7c0b7e919fddf31c8ca8969ea607cbe23d7042c5/credsweeper/utils/util.py#L98
so it would not be spammed during the scan with default (warning) log level

[...]

PR #46, opened 2021-12-08
Less unicode error logging

meanrin:
Fix for #43 

Change `UnicodeError` related logging to info

### DIFF
--- a/credsweeper/utils/util.py
+++ b/credsweeper/utils/util.py
@@ -97,7 +97,7 @@ class Util:
             except UnicodeError:
                 logging.info(f"UnicodeError: Can't read content from \"{path}\" as {encoding}.")
             except Exception as exc:
-                logging.info(f"Unexpected Error: Can't read \"{path}\" as {encoding}. Error message: {exc}")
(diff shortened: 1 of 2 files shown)


Dmitriy-NK (approved):
LGTM

credsweeper/utils/util.py
```diff
@@ -95,7 +95,7 @@ def read_file(cls, path: str, encodings: List[str] = default_encodings) -> List[
                     diff_data = file.read().split("\n")
                 break
             except UnicodeError:
-                logging.warning(f"UnicodeError: Can't read content from \"{path}\" as {encoding}.")
+                logging.info(f"UnicodeError: Can't read content from \"{path}\" as {encoding}.")
             except Exception as exc:
                 logging.error(f"Unexpected Error: Can't read \"{path}\" as {encoding}. Error message: {exc}")
```
ARKAD97, line 100: May be we should add warning only if none of the encodings matches.
And, maybe, add info with specific encoding.

TL; DR: I think we should log once per file, not mutliple logs per every encoding

ARKAD97 (approved):
LGTM

PR #55, opened 2021-12-23
Fix incorrect datatype for np.ndarray objects

Dmitriy-NK:
For some numpy.ndarray, the data type "numpy.array"([function](https://numpy.org/doc/stable/reference/generated/numpy.array.html)) was specified.

### DIFF
--- a/credsweeper/ml_model/features.py
+++ b/credsweeper/ml_model/features.py
@@ -136,11 +136,11 @@ class RenyiEntropy(Feature):
         self.alpha = alpha
         self.norm = norm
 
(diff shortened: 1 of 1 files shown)


meanrin (approved):
`np.array` is a function that converts "array-like" object to `np.ndarray`
https://numpy.org/doc/stable/reference/generated/numpy.array.html#numpy-array

and `np.ndarray` is a class
https://numpy.org/doc/stable/reference/generated/numpy.ndarray.html#numpy.ndarray

csh519 (approved):
LGTM 👀

PR #56, opened 2022-01-11
Faster execution based on regex

meanrin:
**Summary**:
- This PR improves execution time 5-7 times (based on repo under a scan) without any change to the output

Current main branch:
```bash
 time python -m credsweeper --path CredData/data/ -j 16 --save-json dataset_main.json

real    7m34,493s
user    30m57,242s
sys     0m3,632s
```

With this change:
```bash
$ time python -m credsweeper --path CredData/data/ -j 16 --save-json dataset_fast.json

real    1m23,158s
user    5m16,442s
sys     0m3,064s
```

Output is identical:
```bash
$ cmp --silent dataset_main.json dataset_faster.json && echo '### SUCCESS: Files Are Identical! ###' || echo '### WARNING: Files Are Different! ###'
### SUCCESS: Files Are Identical! ###


$ wc -l dataset_main.json && tail -10 dataset_main.json
385218 dataset_main.json
                "value": "b03f5f7f11d50a3a",
                "variable": "PublicKeyToken",
                "entropy_validation": true
            }
        ],
        "ml_probability": null,
        "rule": "Key",
        "severity": "medium"
    }
]

$ wc -l dataset_faster.json && tail -10 dataset_faster.json
385218 dataset_faster.json
                "value": "b03f5f7f11d50a3a",
                "variable": "PublicKeyToken",
                "entropy_validation": true
            }
        ],
        "ml_probability": null,
        "rule": "Key",
        "severity": "medium"
    }

```

-----------------------------------

**Idea**:
Regex engines in python are not very fast. Nearly ~80-90% of CredSweeper execution time spent on single `pattern.search(line)` line
So it would be nicer to find a way to call to regex less
This PR introduce new fields for each rule: `required_substrings` and `min_line_len`
Idea is that if none of lines in `required_substrings` is not present in line under scan, or line length less than `min_line_len` - line would not be regexed under the given rule

`required_substrings` is a substring that is a part of regex itself, and regex would never be matched in line unless this substring is also present
Example:
- For rule api `required_substrings` is `api`
- During execution, api rule would create regex `(?P<variable>(((\'|\\"|`)[^:=\'\\"`<>]*|[^:=\'\\"`<>\\s\\(]*)(?P<keyword>api)[^:=\'\\"`<>]*)(\'|\\"|`)?)\\s*\\]?\\s*(?P<separator>=|:=|:)((\\s|\\w)*\\((\\s|\\w|=|\\()*|\\s*)(?P<value_leftquote>(\\\\)*(b|r)?(\'|\\"|`))?(?P<value>[^\'\\"`(\\\\\\\')(\\\\\\")]{0,1000})(?P<value_rightquote>(\\\\)*(\'|\\"|`))?`
- this regex cannot possible match if line do not have word `api` in it
- we check `if "api" in line`, and if not - no point in running regex at all

`min_line_len` is a minimal possible length that rule require to match it's regex. There is no point in processing line with this rule if number characters is less than this value 
Example:
- Rule `Github Personal Access Token` have regex `(?P<value>ghp_[\w]{36,255})`
- This regex require at least  40 characters to match (`len("ghp_") + 36`)
- If `len(line) < 40` no point in running regex with this line

Exact change list:
- Add `min_line_len` and `required_substrings` for most rules
- `min_line_len` and `required_substrings` can missing from some rules - than this rules would be processed in old way
- Add `Separator.common_as_set` - set of possible separators. Keyword rules require a separator in regex, so if no such character in a line - no point in running regex
- Move `MAX_LINE_LENGTH` to `constants.py`
- Make `scanner.py` use the new field
- Refactor `Scanner.scan` function to make it do less redundant calls

### DIFF
--- a/credsweeper/common/constants.py
+++ b/credsweeper/common/constants.py
@@ -9,6 +9,8 @@ class KeywordPattern:
 
 class Separator:
     common = "=|:=|:"
+    # All unique non-regex characters from `common`
+    common_as_set = "=:"
     c = "="
     java = "="
     json = ":"
@@ -71,3 +73,9 @@ class DiffRowType:
     DELETED = "deleted"
     ADDED_ACCOMPANY = "added_accompany"
     DELETED_ACCOMPANY = "deleted_accompany"
+
+
+MIN_VARIABLE_LENGTH = 1
+MIN_SEPARATOR_LENGTH = 1
+MIN_VALUE_LENGTH = 4
+MAX_LINE_LENGTH = 1500
--- a/credsweeper/rules/config.yaml
+++ b/credsweeper/rules/config.yaml
@@ -6,6 +6,8 @@
   filter_type: GeneralKeyword
   use_ml: true
   validations: []
+  required_substrings:
+    - api
 
 - name: AWS Client ID
   severity: high
@@ -15,6 +17,9 @@
   filter_type: GeneralPattern
   use_ml: true
   validations: []
+  required_substrings:
+    - A
+  min_line_len: 20
 
 - name: AWS Multi
   severity: high
@@ -25,6 +30,10 @@
   filter_type: GeneralPattern
   use_ml: true
   validations: []
+  required_substrings:
+    - AKIA
+    - ASIA
+  min_line_len: 20
 
 - name: AWS MWS Key
   severity: high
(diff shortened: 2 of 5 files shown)


Dmitriy-NK (approved):
Thanks for improvement. The time of analysis is dramatically reduced.

csh519 (approved):
Oh.. regrading `required_substrings`, actually I used on previous version of scanner. 😮 

LGTM 👍

PR #58, opened 2022-01-14
Specify encoding in all `open` function calls

### LINKED ISSUE #57: Default encoding problem
If user use python version with unusual encoding - there can be a problem in reading config files

below you can see a log where user tries to install `pip install .` while having `cp949` encoding as a default
As result `setup.py` cannot properly execute `open("README.md", "r")`

```
Processing /home/user/CredSweeper
  Preparing metadata (setup.py) ... error
  ERROR: Command errored out with exit status 1:
   command: /home/user/CredSweeper/venv_sweeper_meanrin/bin/python -c 'import io, os, sys, setuptools, tokenize; sys.argv[0] = '"'"'/home/user/CredSweeper/setup.py'"'"'; __file__='"'"'/home/user/CredSweeper/setup.py'"'"';f = getattr(tokenize, '"'"'open'"'"', open)(__file__) if os.path.exists(__file__) else io.StringIO('"'"'from setuptools import setup; setup()'"'"');code = f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base /tmp/pip-pip-egg-info-_xn3tpm8
       cwd: /home/user/CredSweeper/
  Complete output (5 lines):
  Traceback (most recent call last):
    File "<string>", line 1, in <module>
    File "/home/user/CredSweeper/setup.py", line 4, in <module>
      long_description = fh.read()
  UnicodeDecodeError: 'cp949' codec can't decode byte 0xe2 in position 6308: illegal multibyte sequence
  ----------------------------------------
WARNING: Discarding file:///home/user/CredSweeper. Command errored out with exit status 1: python setup.py egg_info Check the logs for full command output.
ERROR: Command errored out with exit status 1: python setup.py egg_info Check the logs for full command output.
```

Same issue can happen in other places, such as while reading config files
To fix it it's better to specify encoding in all places where reading is happening

meanrin:
resolve #57 

Changes:
- Specify default encoding as `utf8`
- Put default encoding in all `open` function calls in the project code and tests

`utf8` already was default python encoding on most machines, so this PR would only influence devices with different system encoding

FYI, non-utf8 saved to the output in escaped form, such as `"value": "\u001f\u000e\u00f9t\u00c8\u0001\u00f7"`
That was already default behavior for devices with `utf8` default


csh519 (approved):
Thank you for set default encoding!

LGTM 👍

Dmitriy-NK (approved):
LGTM

PR #62, opened 2022-01-20
Fix for random-order tests

iuriimet:


babenek:
LGTM

babenek (approved):

r-babenko (approved):
Improved tests robustness

meanrin (approved):
This PR adds random order (well, via `pytest-random-order`) so test order would be randomized each time
It would eliminate possibility that our tests somehow dependent on each other (and so on execution order)

In addition this PR adds `deepcopy` to the `tests/rules/test_rule.py`. For some reason `test_create_from_config_p` test fails without `deepcopy` if `pytest-random-order` is in use. This is most likely due to some caching done by `pytest-random-order` that is not very friendly towards dictionaries in fixtures

meanrin:
@iuriimet please conciser adding motivation, as a PR description, to the changes you proposed :) 
It would be easier for other maintainers to understand, especially to the one in different physical locations

iuriimet:
> This PR adds random order (well, via `pytest-random-order`) so test order would be randomized each time It would eliminate possibility that our tests somehow dependent on each other (and so on execution order)
> 
> In addition this PR adds `deepcopy` to the `tests/rules/test_rule.py`. For some reason `test_create_from_config_p` test fails without `deepcopy` if `pytest-random-order` is in use. This is most likely due to some caching done by `pytest-random-order` that is not very friendly towards dictionaries in fixtures

The reason is how python copies compound objects (lists, dictionaries, etc).

A shallow copy constructs a new compound object and then inserts references into it to the objects found in the original.
A deep copy constructs a new compound object and then inserts copies into it of the objects found in the original.

In other words, the 'rule_config' fixture returns a shallow copy of the dictionary with the parameters being tested, and any changes to it will be reflected in the original dictionary

The 'test_create_from_malformed_config_n' test using the @pytest.mark.parametrize decorator sets incorrect parameter values and without deep copying these values will be reflected in the original dictionary and, accordingly, with a random execution order, the 'test_create_from_config_p' test will fail

iuriimet:
> Improved tests robustness

The main goal is to improve the reliability of tests

iuriimet:
> @iuriimet please conciser adding motivation, as a PR description, to the changes you proposed :) It would be easier for other maintainers to understand, especially to the one in different physical locations

Thank you for the comment. I'll try to be more careful :) The main goal is to improve the reliability of tests.

Dmitriy-NK (approved):
Looks good

PR #71, opened 2022-01-28
Remove non-existing attributes from docstring

meanrin:
`MAX_LINE_LENGTH` attribute do not exists in files in `credsweeper/scanner/scan_type/`

`MAX_LINE_LENGTH` do exist, but in the constants:
https://github.com/Samsung/CredSweeper/blob/f82629d5c065e34c99f79035db370d566d58204c/credsweeper/common/constants.py#L81

So it's better to remove this attributes from files in `credsweeper/scanner/scan_type/`, as they are providing a false information


r-babenko (approved):
Ok. Documents are actualized

Dmitriy-NK (commented):
Thanks for updates.

Dmitriy-NK (approved):
LGTM

PR #75, opened 2022-02-09
Integrated pylint checker into workflow

babenek:
Added pylint as checker to CI workflow
Removed unused method
Suppressed extra errors


kmnls (approved):

credsweeper/ml_model/features.py
```diff
@@ -22,6 +22,7 @@ def __call__(self, candidates: List[Candidate]) -> List[bool]:
             candidates: list of candidates to extract features
 
         """
+        # pylint: disable=E1101
```
meanrin, line 25: In this case i think it would be better to add abstract method to the `Feature` class
Similar to 
```python
    @abstractmethod
    def extract(self, candidate: Candidate) -> Any:
        raise NotImplementedError
``` 

In this case while all subclasses do have `extract` in them, it is really missing from the base abstract class, and should be added


We might do it with a different PR, it's just something I want to point out

.github/workflows/test.yml
```diff
@@ -36,4 +36,9 @@ jobs:
         pip install .[ml]
     - name: Test with pytest
       run: |
-        pytest -m pytest --random-order --random-order-bucket=global -m "not api_validation" tests/
+        python -m pytest --random-order --random-order-bucket=global --ignore=docs -m "not api_validation" tests/
+    - name: Analysing the code with pylint
```
meanrin, line 40: Maybe it's better to add a new workflow for the pylint rather than reuse test?
I think motivation might be:

- Tests and pylint can fail separately. In case of they joined into a single task it's not immediately obvious which one did fail
- I suppose there are no need to test pylint on a 3 different OS

babenek: +

credsweeper/ml_model/features.py
babenek: +

meanrin (approved):
LGTM
Pylint would make it easier to check for the code quality

Dmitriy-NK (approved):
Looks good.

PR #80, opened 2022-02-10
Remove `install .` from the test

meanrin:
Remove `pip install .` from test as redundant
`pip install .[ml]` should cover all the base requirements anyway 

In our readme we recommend that user would perform `pip install credsweeper[ml]`, so that would be better to test this exact option. Without additional `pip install .`


Dmitriy-NK (approved):
LGTM

PR #81, opened 2022-02-14
Update value pattern filter description, add link to source

Dmitriy-NK:


babenek (approved):
OK. Let the info will be attached

PR #82, opened 2022-02-14
Wrap imports to use in fuzzing without tensorflow

babenek:
Fuzzing takes too much time for instrumental unused tensorflow and numpy.
Fix naming and imports issues.

### DIFF
--- a/credsweeper/ml_model/ml_validator.py
+++ b/credsweeper/ml_model/ml_validator.py
@@ -33,6 +33,7 @@ class MlValidator:
             raise ModuleNotFoundError(ML_VALIDATOR_IMPORT_ERROR)
         tf.compat.v1.logging.set_verbosity(tf.compat.v1.logging.ERROR)  # To make TF logger quiet
         config = tf.compat.v1.ConfigProto()
+        # pylint: disable=E1101
         config.gpu_options.allow_growth = True  # dynamically grow the memory used on the GPU
         config.log_device_placement = True  # to log device placement (on which device the operation ran)
         sess = tf.compat.v1.Session(config=config)
--- a/credsweeper/ml_model/ml_validator.py
+++ b/credsweeper/ml_model/ml_validator.py
@@ -2,7 +2,7 @@ import json
 import os
 import pathlib
 import string
-from typing import List, Tuple, Union, Any
+from typing import List, Tuple, Union
 
 from credsweeper.common.constants import ThresholdPreset, DEFAULT_ENCODING
 from credsweeper.credentials import Candidate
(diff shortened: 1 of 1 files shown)


credsweeper/ml_model/ml_validator.py
```diff
@@ -2,30 +2,44 @@
...
     from tensorflow.python.keras.preprocessing.sequence import pad_sequences
     from tensorflow.python.keras.utils.np_utils import to_categorical
+    from credsweeper.ml_model import features
+
+    ML_VALIDATOR_IMPORT_ERROR = None
 except ModuleNotFoundError as e:
-    raise ModuleNotFoundError("The ML Validation function cannot be used without additional ML packages.\n"
-                              f"{e.msg}\n"
-                              "Run `pip install credsweeper[ml]` to fix it.")
+    class np:
```
meanrin, line 23: Rather than declaring type to avoid errors in type hints, it's possible to use "forward references"
https://www.python.org/dev/peps/pep-0484/#forward-references
This should still resolve all references related to types, but should not error on execution 

For example by changing
```python
def encode(cls, line, char_to_index) -> np.ndarray:
```
to 
```python
def encode(cls, line, char_to_index) -> "np.ndarray":
```

meanrin (commented):
Can you please describe this change in more details and why this need to be done?

It you don't want to execute/test ml validation module, you may just skip `--ml_validation` CLI flag, 
or not use ml validation if you are using scanner as a library

credsweeper/ml_model/ml_validator.py
babenek: The sources are under instrumental by atheris - it does not know about --ml_validation

meanrin: If atheris can work with `"np.ndarray"` format, it's probably a more clean way to solve this issue, rather than create a new dummy class

For example IDEs (PyCharm) do properly understand this syntax:

babenek: OK

credsweeper/ml_model/ml_validator.py
```diff
@@ -86,35 +91,35 @@ def encode(cls, line, char_to_index) -> np.ndarray:
...
-        return features
+            np_array = np.append(np_array, new_feature)
+        return np_array
 
     @classmethod
-    def extract_unique_features(cls, candidates: List[Candidate]) -> np.ndarray:
+    def extract_unique_features(cls, candidates: List[Candidate]) -> 'np.ndarray':
         """Extract features that can by different between candidates. Join them with or operator."""
-        features = np.array([], dtype=bool)
+        np_array = np.array([], dtype=bool)
```
meanrin, line 109:
```suggestion
        feature_array = np.array([], dtype=bool)
```

`np_array` is a bit too general statement :)
Other than that LGTM, sorry for bothering

meanrin (approved):

iuriimet (approved):
LGTM

PR #84, opened 2022-02-15
Fix suggestions and avoid extra duplicate code

babenek:
Fixed pylint issue "Dangerous default value"
Duplicate code optimized

### DIFF
--- a/credsweeper/scanner/scan_type/pem_key_pattern.py
+++ b/credsweeper/scanner/scan_type/pem_key_pattern.py
@@ -41,15 +41,10 @@ class PemKeyPattern(ScanType):
         """
         assert rule.pattern_type == rule.PEM_KEY_PATTERN, \
             "Rules provided to PemKeyPattern.run should have pattern_type equal to PEM_KEY_PATTERN"
-        line_data = cls.get_line_data(config, line, line_num, file_path, rule.patterns[0], rule.filters)
-
-        if line_data is None:
-            return None
 
         if cls.is_pem_key(lines[line_num:]):
-            candidate = Candidate([line_data], rule.patterns, rule.rule_name, rule.severity, config, rule.validations,
-                                  rule.use_ml)
-            return candidate
+            return cls._get_candidate(config, line, line_num, file_path, rule)
+
         return None
 
     @classmethod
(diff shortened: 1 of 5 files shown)


Dmitriy-NK (dismissed):
No objection on my part. But it's better to do separate PRs for different changes.

meanrin (commented):
Can we also apply `_get_candidate` simplification for the PEM scan?

https://github.com/Samsung/CredSweeper/blob/e63e4a80d1e456297bdd5c3044a654dfa083086f/credsweeper/scanner/scan_type/pem_key_pattern.py#L44

I suppose yes, just need to check `is_pem_key` after `Candidate` would be created

kmnls (approved):
Agree

credsweeper/scanner/scan_type/pem_key_pattern.py
```diff
@@ -41,15 +41,10 @@ def run(cls, config: Config, line: str, line_num: int, file_path: str, rule: Rul
...
-        line_data = cls.get_line_data(config, line, line_num, file_path, rule.patterns[0], rule.filters)
-
-        if line_data is None:
-            return None
 
         if cls.is_pem_key(lines[line_num:]):
-            candidate = Candidate([line_data], rule.patterns, rule.rule_name, rule.severity, config, rule.validations,
-                                  rule.use_ml)
-            return candidate
+            return cls._get_candidate(config, line, line_num, file_path, rule)
```
meanrin, line 46: Other option:

```python
            candidate = cls._get_candidate(config, line, line_num, file_path, rule)
            if candidate is not None and cls.is_pem_key(lines[line_num:]):
                return candidate
            return None
```

I'm not sure, but if I remember correctly `get_line_data` would perform much faster than `is_pem_key`
So this order might be better

I want to test speed in both cases and will post results in an hour

meanrin: It seems that difference between implementations is insignificant and less than 1%
Tested on CredData and few other repos

meanrin (approved):

Dmitriy-NK (approved):
LGTM.

PR #87, opened 2022-02-17
UnitTest enchancement

babenek:
Separate application tests and unit-tests
Skip additional filter for UT

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -41,7 +41,10 @@ jobs:
         pip install .[ml]
     - name: Test with pytest
       run: |
-        python -m pytest --random-order --random-order-bucket=global --ignore=docs -m "not api_validation" tests/
+        # unit-tests
+        python -m pytest --random-order --random-order-bucket=global --ignore=docs --ignore=tests/test_app.py tests/
+        # application test
+        python -m pytest --random-order --random-order-bucket=global --ignore=docs tests/test_app.py
 
 # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # #
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -39,11 +39,11 @@ jobs:
         python -m pip install pytest
         python -m pip install pytest-random-order
         pip install .[ml]
-    - name: Test with pytest
+    - name: UnitTest with pytest
       run: |
-        # unit-tests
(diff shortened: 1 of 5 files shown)


tests/conftest.py
```diff
@@ -59,3 +59,11 @@ def scanner(rule: Rule, config: Config, rule_path: str) -> Scanner:
     scanner = Scanner(config, rule_path)
     scanner.rules = [rule]
     return scanner
+
+
+@pytest.fixture
+def scanner_without_filters(rule: Rule, config: Config, rule_path: str):
```
meanrin, line 65: Some tests do use randomly seeded credentials, so they have a small change to be filtered out by a pattern filter (for example by randomly generating `aaaa` sub-string
This `scanner_without_filters` fixture allows to run regex related tests without filters that can remove some random creds

All other tests still using old `scanner` fixture

.github/workflows/test.yml
```diff
@@ -41,7 +41,10 @@ jobs:
         pip install .[ml]
     - name: Test with pytest
       run: |
-        python -m pytest --random-order --random-order-bucket=global --ignore=docs -m "not api_validation" tests/
+        # unit-tests
+        python -m pytest --random-order --random-order-bucket=global --ignore=docs --ignore=tests/test_app.py tests/
+        # application test
+        python -m pytest --random-order --random-order-bucket=global --ignore=docs tests/test_app.py
```
meanrin, line 47: I assume that new test command can be separated to a new step in the `test.yml`
To add clearer divide between unit and integration tests, if we want to 

Btw, I don't see direct benefits dividing tests into 2 separate commands, they are still both just a pytest tests anyway

meanrin, line 47: I assume that new test command can be separated to a new step in the `test.yml`

babenek: Unit tests might be used for code coverage. Separating steps lead to overhead in workflow - double python setup.

meanrin: I meant something like

```yaml
- name: Unit tests
      run: ...
- name: Application tests
      run: ... 
```

That should not add any overhead, just additional separation between different types of tests 

image

tests/conftest.py
babenek: UnitTests should avoid any "chance of".

.github/workflows/test.yml
babenek: ok

kmnls (approved):

iuriimet (approved):
Agree

PR #90, opened 2022-03-01
Fix incorrect type declarations in util

cuspymd:
There are a lot of `mypy` type errors in the source codes.
I fixed incorrect type declarations in util.

### DIFF
--- a/credsweeper/utils/util.py
+++ b/credsweeper/utils/util.py
@@ -3,11 +3,20 @@ import math
 import os
 from dataclasses import dataclass
 from typing import Dict, List, Tuple
+from typing_extensions import TypedDict
 
 import whatthepatch
 from regex import regex
 
 from credsweeper.common.constants import Chars, DiffRowType, KeywordPattern, Separator
+
+
+DiffDict = TypedDict("DiffDict", {
+    "old": int,
+    "new": int,
+    "line": str,
+    "hunk": str
+})
 
 
 @dataclass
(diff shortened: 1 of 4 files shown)


credsweeper/utils/util.py
```diff
@@ -30,7 +39,7 @@ def get_extension(cls, file_path: str) -> str:
         return extension
 
     @classmethod
-    def get_keyword_pattern(cls, keyword: str, separator: Separator = Separator.common) -> regex.Pattern:
+    def get_keyword_pattern(cls, keyword: str, separator: str = Separator.common) -> regex.Pattern:
```
meanrin, line 42: Hm, for some reason I thought that  `Separator` is a Enum, not just class with constants

credsweeper/utils/util.py
```diff
@@ -160,12 +169,12 @@ def preprocess_file_diff(cls, changes: List[Dict]) -> List[DiffRowData]:
         for change in changes:
             if change.get("old") is None:
                 # indicates line was inserted
-                rows_data.append(DiffRowData(DiffRowType.ADDED, change.get("new"), change.get("line")))
```
meanrin, line 163: `.get` here allows to pass `None` in case of a missing value, preventing `KeyError` exception 
https://stackoverflow.com/a/11041421

I'm not sure that raising `KeyError` here is necessary or not
@Dmitriy-NK , what do you think?

meanrin (commented):
Sorry for a long response time
Mostly LGTM, thanks a lot!

But I'm not sure about changing `.get(` to `[` in `util.py`

@Dmitriy-NK , can you please comment on the question below?
Are we okay with possible `None`s as function arguments, or is it better to raise a `KeyError` ?

Dmitriy-NK (commented):
No objection on my part.

babenek (approved):
Looks Good. 
May be merged

meanrin (approved):

PR #72, opened 2022-02-03
Limit TF version by 2.8.0

meanrin:
Newly released TF 2.8.0 have a different Keras API

Until new API implemented in the code, this PR limit the TF version to `<2.8.0`

```
$ python -m credsweeper --path tests/samples/password --ml_validation
Traceback (most recent call last):
  File "/home/user/share/code/credsweeper_replicas/meanrin/CredSweeper/credsweeper/ml_model/ml_validator.py", line 12, in <module>
    from tensorflow.python.keras.preprocessing.sequence import pad_sequences
ModuleNotFoundError: No module named 'tensorflow.python.keras.preprocessing'
```

image

PR #93, opened 2022-03-09
Add if to support new keras imports in TF 2.8.0

meanrin:
Related to: https://github.com/Samsung/CredSweeper/pull/72 

Due to the change in keras library location in tf 2.8.0:
- Add if to check tf version
- Add different imports for 2.8 and prior versions


babenek (approved):
LGTM. Fixes security issues

PR #106, opened 2022-04-29
MlValidator object is created once

babenek:
Speedup when Credsweeper object is created and used many times with ml_validation - skip multiple model load

### DIFF
--- a/credsweeper/app.py
+++ b/credsweeper/app.py
@@ -67,6 +67,7 @@ class CredSweeper:
         config_dict["use_filters"] = use_filters
         config_dict["find_by_ext"] = find_by_ext
 
+        self.ml_validator = None
         self.config = Config(config_dict)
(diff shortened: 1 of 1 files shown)


csh519 (commented):
Good point :eyes: 

But is there a case where ML models are created and executed many times?
I think it is same result in normal case.

babenek:
During the model is created, there is heaviest process - model file loading.

csh519:
> During the model is created, there is heaviest process - model file loading.

Yes, I know. But in this code `post_processing()` method just called once in normal execution scenario.. isn't it?

babenek:
> 

in APP case - yes, but in API using - there might be variations...

csh519 (approved):
Thank you for explanation!

LGTM :+1:

Dmitriy-NK (approved):
No objection on my part.

PR #107, opened 2022-05-03
Dynamic testing in github actions with atheris framework

babenek:
Fuzzing with atheris framework with 25 corpuses
+ unit test for #106


csh519:
I have a question, it seems there are more than 4k corpus files.
How are these files generated? 
And is it need to update when the files(CredSweeper) are updated?

babenek:
Corpus files are generated by atheris framework when it enabled. 
The files (4K) were generated with more than 10M cycles of fuzzing and now it allow spend much less time to cover 74% of code.
It means, that when CredSweeper is refactored deeply - the corpus files do not make the same coverage. 
That should be a notifier, an indicator...
Actually, libfuzzer should minimize the files if duplicate... I suspend the PR till investigation.

babenek:
Created corpus minimization scripts and filtered only 25 corpus with 74% coverage

.github/workflows/fuzz.yml
```diff
@@ -0,0 +1,173 @@
...
+        DESCRIPTION="..."
+        for stage in 'coverage' 'new corpus' 'crash corpus'; do
+          CONTEXT="Fuzzing / ${stage}"
+          #dbg
+          echo "CONTEXT='$CONTEXT'"
+          curl -v -i \
+            --request POST \
+            --header "Accept: application/vnd.github.v3+json" \
+            --header "Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}" \
+            "https://api.github.com/repos/${GITHUB_REPOSITORY}/statuses/${GIT_COMMIT}" \
```
babenek, line 46:
```suggestion
            https://api.github.com/repos/${GITHUB_REPOSITORY}/statuses/${GIT_COMMIT} \
```

tests/test_main.py
```diff
@@ -7,7 +7,7 @@
 
 import pytest
 
-from credsweeper import __main__, ByteContentProvider, StringContentProvider
+from credsweeper import __main__, ByteContentProvider, StringContentProvider, file_handler
```
csh519, line 10: It would be good to change with below code.

``` python
from credsweeper.file_handler.byte_content_provider import ByteContentProvider
```

It isn't match with `TextContentProvider`. Please check.

csh519: Or you can merge both of them like this.

``` python
from credsweeper.file_handler import ByteContentProvider, TextContentProvider
```

csh519 (changes requested):

tests/test_main.py
babenek: Ok. The unit test is not related for fuzzing - i'll move it to other PR

fuzz/fuzzing.sh
```diff
@@ -0,0 +1,22 @@
...
+    -seed=$(date +%s) \
+    -atheris_runs=$(( 100500 + $(ls corpus | wc -l) )) \
+    -verbosity=1 \
+    corpus/ \
+    ;
+
+
+
+#    -mutation_graph_file=mutation_graph_file.txt \
+#    -print_full_coverage=1 \
```
csh519, line 22: It seems useless lines. Could you remove those lines?

csh519 (commented):
There are some files that don't have any descriptions.. `fuzz/__main.py__`, `fuzz/fuzzing.sh`, `fuzz/merge.sh`, 

Please add descriptions by creating `README.md` file for `fuzz` directory or any other ways are good.

kmnls (approved):

Dmitriy-NK (approved):
Thanks for updates

csh519 (approved):
There is no descriptions for `__main__.py` but It seems `__main__.py` is used by `atheris`.
So i think it would be good to merge as it is.
Thank you for add `atheris` framework.

LGTM :+1:

PR #108, opened 2022-05-05
Perform CodeQL Analysis

babenek:
by github


kmnls (changes requested):
https://codeql.github.com/

It can't be used for automated analysis, continuous integration or continuous delivery, whether as part of normal software engineering processes or otherwise, except in the express cases set forth herein. For these uses, [contact the sales team.](https://enterprise.github.com/contact)

PR #109, opened 2022-05-05
Create SECURITY.md

babenek:
File to describe which versions are supported


csh519 (approved):
Thank you for adding `SEUCIRTY.md` file.
I think it would be nice if there was more content.

LGTM :+1:

kmnls (approved):

PR #110, opened 2022-05-06
UnitTest for PR#106

babenek:
UnitTest for https://github.com/Samsung/CredSweeper/pull/106


csh519 (commented):
I can't understand why it is a unit test of #106.
It seems read files by `ByteContentProvider` and just scan them and it would be good to set the test path to `credsweeper/tests/samples`.

kmnls (dismissed):

csh519:
@babenek 
Thank you for comment!
But still i couldn't understand fully..

1. I think the purpose of test is check whether `CredSweeper` create ml model once. So I think check attribute id that from module like below or some other ways.

``` python
>>> yapf.__getattribute__
<method-wrapper '__getattribute__' of module object at 0x7f4f012a45e0>
```

2. Regrading real path using, could you explain it more? It seems just scan all `CredSweeper` directory now, but I think only `tests/samples` directory should used for this case. Because I think it is enough to test ml model working well.

babenek:
@csh519 
fixed: 
- use id() for checking ml_validator. More robustness may be when ml model file is removed after first initiation.
- use 'samples' dir for the test - there should be constant values in candidates.

tests/utils/test_util.py
```diff
@@ -0,0 +1,174 @@
...
+        n = 65536
+        while 0 < n:
+            try:
+                unicode_char = chr(random.randint(0, 0x10FFFF))
+                encoded_bin = unicode_char.encode('utf-16-be')
+                utf16_char = encoded_bin.decode('utf-16-be')
+                if unicode_char != utf16_char:
+                    raise Exception(f"Wrong refurbish:{unicode_char} {encoded_bin} {utf16_char}")
+            except Exception as exc:
+                # print(f'{exc}')
```
Dmitriy-NK, line 156: Please update line if necessary.

credsweeper/utils/util.py
```diff
@@ -108,6 +108,37 @@ def read_file(cls, path: str, encodings: Tuple[str, ...] = default_encodings) ->
...
+            list of file rows in a suitable encoding from "encodings",
+            if none of the encodings match, an empty list will be returned
+
+        """
+        lines = []
+        for encoding in encodings:
+            try:
+                text = content.decode(encoding)
+                if content != text.encode(encoding):
+                    raise UnicodeError
```
iuriimet, line 132: it may be better to use chardet (https://pypi.org/project/chardet/) package to detect the encoding

tests/utils/test_util.py
babenek, line 156:
```suggestion
```

tests/utils/test_util.py
```diff
@@ -0,0 +1,173 @@
...
+        while 0 < n:
+            try:
+                unicode_char = chr(random.randint(0, 0x10FFFF))
+                encoded_bin = unicode_char.encode('utf-16-le')
+                utf16_char = encoded_bin.decode('utf-16-le')
+                if unicode_char != utf16_char:
+                    # print(f"Wrong refurbish:{unicode_char} {encoded_bin} {utf16_char}")
+                    continue
+            except Exception as exc:
+                # print(f'{exc}')
```
babenek, line 125:
```suggestion
```

Dmitriy-NK (approved):
Thanks for updates.

iuriimet (approved):
Loosk good

credsweeper/utils/util.py
iuriimet: I have tested the 'chardet' package (https://github.com/iuriimet/CredSweeper/tree/test_chardet).
In many cases the encoding is detected incorrectly

PR #118, opened 2022-05-18
Install credsweeper as cli tool additionally

babenek:
Credsweeper is installed as utility in system (or virtual environment) additionally with 
```
pip install credsweeper
```
then it might be launched in console like
```
$ credsweeper --path .
```


meanrin (approved):
LGTM
Will you update Readme and Docs with a different PR?

PR #120, opened 2022-05-19
Using CredSweeper as github action

babenek:
- added files to use the project as step in github workflow as [Docker container action](https://docs.github.com/en/actions/creating-actions/creating-a-docker-container-action)
- test in #121 
- refactored workflows code
- TODO: rename 'build' to 'test' in requires checks

### DIFF
--- a/.github/workflows/fuzz.yml
+++ b/.github/workflows/fuzz.yml
@@ -1,5 +1,6 @@
-# This workflow will install Python dependencies, run tests and lint with a variety of Python versions
-# For more information see: https://help.github.com/actions/language-and-framework-guides/using-python-with-github-actions
+# This workflow tests the project with atheris fuzzing framework
+# Coverage should not be less than with precommitted corpuses
+# Otherweise fuzzing workaround has to be performed and new corpuses are committed
 
 name: Fuzzing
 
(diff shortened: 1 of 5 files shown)


docs/howto/how-to-contribute.md
```diff
@@ -24,6 +24,7 @@ You can also look at our [issues](https://github.com/Samsung/CredSweeper/issues)
 
 For small changes, you can use GitHub's web interface. Simply click the **Edit the file in your fork of this project** on the file you'd like to change.
 GitHub creates the new branch for you when you submit the changes.
+VCS(git) requirement: the branch MUST be forked after latest release.
```
meanrin, line 27: Can you please comment on why is that?
Can't I just fetch main branch after the release?

git_workflow.sh
```diff
@@ -0,0 +1,62 @@
+#!/bin/bash
```
meanrin, line 1: Some projects use separate directory for workflows, such as here: https://github.com/SAP-samples/datahub-integration-examples/tree/main/GitWorkflow/scripts 
What do you think about it? 
It's probably unnecessary if we have only a one script, but it might be useful if we would add more in the future

meanrin (approved):

git_workflow.sh
csh519: I also think it's good to separate the scripts.

babenek: It was already merged in #119.
May be with next workflow

kmnls (approved):

csh519 (commented):
It seems there are CI issues.. Could you please check github checks?

meanrin (approved):
LGTM

Issue #124, opened 2022-05-24
Need to ignore files with large size


alf0505:
It would be great if I can have ignore option not to scan large files which take too much time.

csh519:
Hello!

I also think set limitation of file is good.
I'll make PR regarding this.

Thank you for suggestion.

PR #125, opened 2022-05-24
Add size limit option

csh519:
I added size limit option regrading #124.

And here is change list:
- Add `size_limit` option. (eg. 1GB / 10 MiB/ 1024)
- Unify option style. (eg. `--save-json` to `--save_json`)
- Update `guide.rst` file.
- Clean code.

### DIFF
--- a/credsweeper/__main__.py
+++ b/credsweeper/__main__.py
@@ -42,7 +42,7 @@ def get_arguments() -> Namespace:
                         default=None,
                         dest="rule_path",
                         metavar="PATH")
-    parser.add_argument("--find-by-ext",
+    parser.add_argument("--find_by_ext",
                         help="find files by predefined extension.",
                         dest="find_by_ext",
                         action="store_true")
@@ -84,7 +84,7 @@ def get_arguments() -> Namespace:
                         help="parse .gitignore files and skip credentials from ignored objects",
                         dest="skip_ignored",
                         action="store_true")
-    parser.add_argument("--save-json",
+    parser.add_argument("--save_json",
                         nargs="?",
                         help="save result to json file (default: output.json)",
                         const="output.json",
@@ -105,7 +105,7 @@ def get_arguments() -> Namespace:
                         metavar="POSITIVE_INT")
     parser.add_argument("--version",
                         "-V",
-                        help="Show program's version number and exit",
+                        help="show program's version number and exit",
                         action="version",
                         version=f"CredSweeper {__version__}")
     return parser.parse_args()
--- a/credsweeper/__main__.py
+++ b/credsweeper/__main__.py
@@ -98,11 +98,9 @@ def get_arguments() -> Namespace:
                         metavar="LOG_LEVEL",
                         choices=list(Logger.LEVELS))
     parser.add_argument("--size_limit",
-                        help="set size limit of files that for scanning (unit: MiB)",
-                        type=positive_int,
+                        help="set size limit of files that for scanning (eg. 1GB / 10MiB / 1000)",
                         dest="size_limit",
-                        default=None,
-                        metavar="POSITIVE_INT")
+                        default=None)
     parser.add_argument("--version",
                         "-V",
                         help="show program's version number and exit",
--- a/credsweeper/__main__.py
+++ b/credsweeper/__main__.py
@@ -98,7 +98,7 @@ def get_arguments() -> Namespace:
                         metavar="LOG_LEVEL",
                         choices=list(Logger.LEVELS))
     parser.add_argument("--size_limit",
-                        help="set size limit of files that for scanning (unit: MB)",
+                        help="set size limit of files that for scanning (unit: MiB)",
                         type=positive_int,
                         dest="size_limit",
                         default=None,
--- a/credsweeper/__main__.py
+++ b/credsweeper/__main__.py
@@ -42,7 +42,7 @@ def get_arguments() -> Namespace:
                         default=None,
                         dest="rule_path",
                         metavar="PATH")
-    parser.add_argument("--find_by_ext",
+    parser.add_argument("--find-by-ext",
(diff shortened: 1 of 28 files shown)


credsweeper/__main__.py
```diff
@@ -84,7 +84,7 @@ def get_arguments() -> Namespace:
                         help="parse .gitignore files and skip credentials from ignored objects",
                         dest="skip_ignored",
                         action="store_true")
-    parser.add_argument("--save-json",
+    parser.add_argument("--save_json",
```
meanrin, line 87: Let's change syntax in a different PR, as this change will break all scripts that using scanner CLI for sure :(
Alternatively, we might make both options supported, so both `save_json` and legacy `save-json` would work, [example](https://stackoverflow.com/a/11155124)

credsweeper/__main__.py
```diff
@@ -97,9 +97,15 @@ def get_arguments() -> Namespace:
                         dest="log",
                         metavar="LOG_LEVEL",
                         choices=list(Logger.LEVELS))
+    parser.add_argument("--size_limit",
+                        help="set size limit of files that for scanning (unit: MB)",
```
meanrin, line 101: Small notation issue, but we are probably talking about `MiB`, not `MB`

Apparently there is a difference between Megabyte (MB, 1 000 000 bytes) and [Mebibyte](https://simple.wikipedia.org/wiki/Mebibyte) (MiB, 1 048 576 bytes)
Some systems got it confuse, so that's why you may have noticed that file size may change if you upload a file from Windows to Linux or vice versa

This isn't really a big deal, as almost all the software has confused it anyway. I just want to point this out.

meanrin (changes requested):
Feature itself is great. However, lets not change CLI args, as we are risking to break scanner for many people

credsweeper/__main__.py
csh519: Good point I'll make other PR for this.

credsweeper/__main__.py
csh519: Fixed with [62c3f76](https://github.com/Samsung/CredSweeper/pull/125/commits/62c3f76463fdd742b79829cd8001e1d232c9021e).

Thank you for review.

credsweeper/__main__.py
csh519: Revoked changes with [858ff7a](https://github.com/Samsung/CredSweeper/pull/125/commits/858ff7ac8d1c45bb635b61925fda0d9827a9e2e6).

meanrin (approved):
LGTM, thank you for the quick response

credsweeper/file_handler/file_path_extractor.py
```diff
@@ -105,3 +106,12 @@ def check_exclude_file(cls, config: Config, path: str) -> bool:
...
         return False
+
+    @classmethod
+    def check_file_size(cls, config: Config, path: str) -> bool:
+        if config.size_limit is None:
+            return False
+        if os.path.getsize(path) > config.size_limit * 1024 * 1024:  # Convert size_limit to byte
+            return True
+        else:
+            return False
```
babenek, line 117: The line is not covered with a unit test.

credsweeper/file_handler/file_path_extractor.py
```diff
@@ -105,3 +106,12 @@ def check_exclude_file(cls, config: Config, path: str) -> bool:
         if Util.get_extension(path) in config.exclude_extensions:
             return True
         return False
+
+    @classmethod
+    def check_file_size(cls, config: Config, path: str) -> bool:
+        if config.size_limit is None:
+            return False
+        if os.path.getsize(path) > config.size_limit * 1024 * 1024:  # Convert size_limit to byte
```
babenek, line 114: may be use human readable arguments 1M 2Gb and so on?

babenek, line 114: may be use human readable arguments 1M 2Gb and so on?

babenek (commented):
Please consider whether a library (e.g. https://github.com/xolox/python-humanfriendly) may be used for human readable arguments. By default - in bytes.

babenek:
PR description might be outdated

credsweeper/file_handler/file_path_extractor.py
```diff
@@ -105,3 +106,12 @@ def check_exclude_file(cls, config: Config, path: str) -> bool:
         if Util.get_extension(path) in config.exclude_extensions:
             return True
         return False
+
+    @classmethod
+    def check_file_size(cls, config: Config, path: str) -> bool:
+        if config.size_limit is None:
```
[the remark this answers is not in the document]
csh519: I attempted to distinguish between user input and default value.
If user entered 0 as input, all files that have size more than 0 are should be filtered.
And if user entered nothing, `size_limit` is `None` and it means no more filtering the files.
So the two are different.

credsweeper/file_handler/file_path_extractor.py
csh519: I covered them with f35951f.

Thank you for review.

credsweeper/file_handler/file_path_extractor.py
csh519: I added `humanfriendly` module for file size with [bf433d1](https://github.com/Samsung/CredSweeper/pull/125/commits/bf433d1977b7ecb73222f1e52aa99f864c3e8312).

Thank you for review.

babenek (approved):
Works perfect! Hope, feature will be not misused.
Pity, setup.py was reformatted by strict yapf. :(

meanrin (dismissed):

babenek (approved):
LGTM.
In case of wrong size input it throws humanfriendly.InvalidSize - should be enough to understand the value.

meanrin (approved):

PR #128, opened 2022-05-26
Fix git workflow issue

### LINKED ISSUE #127: VCS checks fail on main branch
[VCS checks](https://github.com/Samsung/CredSweeper/blob/881b688d96f2dca3e8b60c29d4b244a6e7d93ad9/.github/workflows/vcs.yml) is continuously failing.

@babenek could you please check this issue?

babenek:
fixes #127

### DIFF
--- a/.github/workflows/vcs.yml
+++ b/.github/workflows/vcs.yml
@@ -31,7 +31,11 @@ jobs:
 
     - name: Get latest release tag name
       run: |
-        API_RELEASE_URL=$(echo "${{ github.event.pull_request.base.repo.releases_url }}")
+        if [ "pull_request" == "${{ github.event_name }}" ]; then
+            API_RELEASE_URL=$(echo "${{ github.event.pull_request.base.repo.releases_url }}")
+        else
+            API_RELEASE_URL=$(echo "${{ github.event.repository.releases_url }}")
+        fi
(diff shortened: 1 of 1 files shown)


.github/workflows/vcs.yml
```diff
@@ -26,7 +26,11 @@ jobs:
 
     - name: Get latest release tag name
       run: |
-        API_RELEASE_URL=$(echo "${{ github.event.pull_request.base.repo.releases_url }}")
+        if [ "pull_request" == "${{ github.event_name }}" ]; then
+            API_RELEASE_URL=$(echo "${{ github.event.pull_request.base.repo.releases_url }}")
+        else
+            API_RELEASE_URL=$(echo "${{ github.event.repository.releases_url }}")
```
csh519, line 32: I don't know well about it has different value when main branch and PR status.
But if they're have same value how about remove if condition and just use `github.event.repository.release_url`?

babenek: The config is different for type pull_request and push. It was discovered with debug step https://github.com/babenek/CredSweeper/commit/8848aa46268b3092554c8ab6740ac3e21c210c67

kmnls (approved):

csh519 (approved):
Thank you for fix issue.

LGTM :+1:

PR #129, opened 2022-05-27
Documentation check workflow

babenek:
- documentation build step involved in common workflow. Fixed one warning
- Pylint checks  missed docstrings. All missed docstrings are skipped at current positions.
They will raise error in future. Preferred way - fix it.
- Git workflow script was moved to separate folder.


.github/workflows/test.yml
```diff
@@ -262,4 +262,10 @@ jobs:
     - name: Analysing the code with mypy and minimum Python version 3.9
       run: mypy --config-file .mypy.ini --python-version=3.9 --module credsweeper | sort
 
+    - name: Analysing the code with pylint for NEW missed docstrings of classes or functions
+      run: pylint --disable=R,W,C0114,C0103,C0415,C0200,C0325 --verbose credsweeper 2>/dev/null | grep '^credsweeper/' | LC_ALL=C sort -g | diff cicd/missed_docstrings.txt -
```
csh519, line 266: In my opinion it is not fully guarantee that the document has been updated.
Is there any other way?

How about compare the diff full documents? It just my thought.

babenek: Yes, the check might detect new lines without docstring and known lines it they moved.
In case when old lines are kept on old places - they do not raise error.
So, properly direction to fix missed docstring partially is from tail to head.

kmnls (approved):

csh519 (approved):
Thank you for add missed doc string check.

LGTM 👍

PR #130, opened 2022-05-31
Fix limit for dynamic testing to 74%

babenek:
The marker might indicate for valuable changes (big refactoring). Then, re-fuzzing step should be done.


csh519 (commented):
I don't know the purpose of this PR. Could you explain further?

babenek:
The workflow is "Dynamic code analysis" which is a requirement of https://bestpractices.coreinfrastructure.org/en/projects/6055#analysis.
So, the PR freezes condition of fuzzed coverage for 74%. 
There were no assertions. Only crashes(if they appeared) are collected now.

csh519 (approved):
Thank you for explanation.

LGTM :+1:

PR #133, opened 2022-06-03
Format and unvaluable changes

babenek:
- Fixed code style for text files
- Enabled test for PyPi pre-build on main branch fetch
- Balanced github actions to reduce payload for check-out and package installation.
- 'release_test' - when the application and required packages are installed with setup.py with latest versions
- 'development_test' - all packages are installed from requirements.txt with fixed versions

When the PR will be approved - required checks have to be updated.


.github/workflows/check.yml
```diff
@@ -0,0 +1,165 @@
+# This workflow performs static analysis and checks coding style
```
meanrin, line 1: What do you think about renaming this file to `style.yml`, `style_check.yml` or something else that would indicate that this is about codding style and not actual tests?

.github/workflows/pypi.yml
```diff
@@ -25,10 +27,21 @@ jobs:
...
-        python setup.py sdist bdist_wheel
+      run:
         twine upload dist/*
+
+    - name: Test for creation sphinx documentations
+      if: ${{ 'release' != github.event_name }}
+      run: |
+        cd docs
+        pip install -r requirements.txt
+        make html
```
meanrin, line 47: Do we have a negative example for this test? What will happen if `sphinx` returns error?

credsweeper/common/keyword_checklist.txt
```diff
@@ -942,4 +942,4 @@ yellow
 yield
 zeros
 .json
-.xml
\ No newline at end of file
+.xml
```
meanrin, line 945: Maybe I'm blind but I cannot see where is a difference :D
I assume it's related to a newline but git not showing it

docs/source/conf.py
```diff
@@ -78,12 +78,11 @@
...
 html_static_path = ['_static']
 
-
 html_theme = 'sphinx_rtd_theme'
 
 html_theme_options = {
-    'logo_only': True,
-    'navigation_depth': 3
+    'logo_only': True,  #
+    'navigation_depth': 3  #
```
meanrin, line 85: What is the purpose of these comments?

babenek: to keep the lines separately - YAPF merges them

.github/workflows/pypi.yml
babenek: In previous PR i added default key -W to fail in warning case

.github/workflows/check.yml
babenek: There are multiple check types.

babenek: It is a common name i suppose.

csh519 (approved):
Thank you for clean workflow files and formatting. :broom: 

LGTM :+1:

kmnls (approved):

iuriimet (approved):

PR #134, opened 2022-06-07
Fix sphinx documentation workflow

babenek:
- used 'hexdump' tool instead python to simplify workflow
- fix sphinx build with install all packages
- enable pypi workflow for PR but without publishing

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -37,7 +37,7 @@ jobs:
                 lastbyte=$(hexdump -v -e '/1 "%02X\n"' ${f} | tail -1)
                 echo "Last byte is '${lastbyte}'"
                 if [ "0A" != "${lastbyte}" ]; then
-                    echo "File ${f} has inappropriated line ending"
+                    echo "File ${f} has inappropriate line ending"
                     tail -1 ${f} | hexdump -C
                 else
                     n=$(( ${n} - 1 ))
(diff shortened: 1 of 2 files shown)


csh519:
@babenek 
Did you checked this error? https://github.com/Samsung/CredSweeper/runs/6773957745?check_suite_focus=true
It seems `regex` module required in `docs/requiremnets.txt` file.

babenek:
@csh519
The PR fixes the issue and checks it every PR instead fetch.
regex and others packages requires due they are imported in credsweeper.
The solution is to install all required packages for credsweeper and then - packages for sphinx

.github/workflows/check.yml
```diff
@@ -34,10 +34,11 @@ jobs:
...
             if echo "${filetype}" | grep -q '.*text.*'; then
                 echo "CHECK:'${filetype}'"
-                x=$(python -c "import sys, os; f=open('${f}', 'rb'); f.seek(-1, os.SEEK_END); print(format(ord(f.read()),'02X')); f.close();")
-                if [ "0A" != "${x}" ]; then
-                    echo "File ${f} has unappropriated line ending '${x}'"
-                    tail -1 ${f}
+                lastbyte=$(hexdump -v -e '/1 "%02X\n"' ${f} | tail -1)
+                echo "Last byte is '${lastbyte}'"
+                if [ "0A" != "${lastbyte}" ]; then
+                    echo "File ${f} has inappropriated line ending"
```
meanrin, line 40:
```suggestion
                    echo "File ${f} has inappropriate line ending"
```
Possible typo

iuriimet (approved):

PR #135, opened 2022-06-08
Some docstring and warnings were fixed

babenek:


csh519 (approved):
Thank you for fix docstring and warnings.

LGTM :+1:

credsweeper/credentials/line_data.py
```diff
@@ -139,7 +139,7 @@ def set_pattern_match_groups(self) -> None:
         if match_obj is None:
             return
 
-        def get_group_from_match_obj(match_obj: regex.Match, group: str) -> Optional[str]:
+        def get_group_from_match_obj(match_obj: regex.Match, group: str) -> Any:
```
meanrin, line 142: Why not `Optional[str]`?
https://stackoverflow.com/a/39429578

credsweeper/file_handler/content_provider.py
```diff
@@ -9,6 +9,9 @@ class ContentProvider(ABC):
 
     @abstractmethod
     def __init__(self, file_path: str, change_type: Optional[str] = None, diff: Optional[List[Dict]] = None) -> None:
+        self.file_path = file_path
+        self.change_type = change_type
+        self.diff = diff
         raise NotImplementedError()
```
meanrin, line 15: But this function raises an error
Why even assign anything?

kmnls (approved):

credsweeper/credentials/line_data.py
babenek: credsweeper/credentials/line_data.py:144: error: Returning Any from function declared to return "Optional[str]"
it is not 'str' but AnyStr

credsweeper/file_handler/content_provider.py
babenek: many IDE warnings which cannot find the attributes in the base class

credsweeper/credentials/line_data.py
meanrin: So why not to specify `AnyStr`? 
[AnyStr](https://docs.python.org/3/library/typing.html#typing.AnyStr)

`Any` is just too generic to provide any meaningful information

credsweeper/file_handler/content_provider.py
meanrin: > many IDE warnings which cannot find the attributes in the base class

However, adding this lines would decrease test coverage, as now this method interpreted as a real code by linters, not just as placeholder

If this constructor is never used (which it seems) we can just remove it, why keep it? 

Other option is to declare properties as functions:
https://stackoverflow.com/a/2736417

meanrin:
@babenek please do not merge if there are unresolved conversations left 
Personally I don't think that changes in [credsweeper/file_handler/content_provider.py](https://github.com/Samsung/CredSweeper/pull/135/files/b3214700c94485568739ff5763165c0c89b91b10#diff-5da5e0a9b4493c60bfb6bf4036bd33ceef79310219bf4633adcc9e91cfa85879) are appropriate to merge

PR #141, opened 2022-06-16
bugfix: wrong response from server

babenek:
Possibly uncaught crash when server replies unexpected data


kmnls (approved):

iuriimet (approved):
Agree

Dmitriy-NK (approved):
Looks good

PR #143, opened 2022-06-20
Bugfix SLACK api validation

babenek:
Uncaught exception might be happen in case wrong server response.

### DIFF
--- a/credsweeper/validations/slack_token_validation.py
+++ b/credsweeper/validations/slack_token_validation.py
@@ -1,4 +1,3 @@
-import logging
 from typing import List
 
 import requests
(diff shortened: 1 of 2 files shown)


csh519 (commented):
Could you explain more about error?
If there are exception cases that we didn't covered, we should add them, not just make them pass.

credsweeper/validations/slack_token_validation.py
```diff
@@ -1,3 +1,4 @@
+import logging
```
csh519, line 1: Flake8 detected `import logging` is unused.
Please check.

csh519: Thank you for explanation.

csh519 (approved):
Maybe `except Exception:` can be replace with `finally:`. But it is okay.

LGTM :+1:

kmnls (approved):

PR #144, opened 2022-06-22
Fuzzing with server response mocking

babenek:
* fuzzing with server response mocking
* reordered consumers and minimized bytes of fuzz data (Weird order of bytes in data provider https://github.com/google/atheris/pull/35)
* fixed possibly issues with wrong respone


kmnls (approved):
Agree with changes in fuzzing

kmnls (approved):

csh519 (approved):
Approve fuzz changes.

Issue #126, opened 2022-05-24
Enable ml_validation by default


babenek:
Side effect of binary files scanning is many false-positive cases without --ml_validation.
Users might use minimal set of arguments (e.g. '--path' only). 
It leads to big false-positive warnings due binary files are decoded like Latin1 text. 
Evidently, random ASCII bytes look like a good password. Thus '--ml_validation' filters many of the warnings.

Enabling the option supposes to install ML library by default.

csh519:
As we discussed, I'll create a PR that defaults to `ml_validation` option.
It might be done before next week.

PR #146, opened 2022-06-27
Change --ml_validation option by default

csh519:
Changed `--ml_validation` option by default. #126

**Changes**
- Remove `--ml_validation` argument.
- Change test cases, yaml files, and so on.
- Fix typo.

**Usage**
It is default to use ML Validation with medium(`ThresholdPreset`).
``` bash
python -m credsweeper --path ~/test/
```
If you want to skip ML Validation, you can use `--ml_threshold 0` option.
``` bash
python -m credsweeper --path ~/test/ --ml_threshold 0
```


credsweeper/app.py
```diff
@@ -76,10 +73,15 @@ def __init__(self,
         self.json_filename: Optional[str] = json_filename
         self.ml_batch_size = ml_batch_size
         self.ml_threshold = ml_threshold
-        if ml_validation:
+        if self._use_ml_validation():
```
babenek, line 76: As soon we make ML as main feature - import should be used as usually. And setup.py ...

experiment/src/prepare_data.py
```diff
@@ -7,8 +7,8 @@ def execute_scanner(dataset_location: str, result_location_str, j, use_ml=False)
     """Execute CredSweeper as a separate process to make sure no global states is shared with training script"""
     dir_path = os.path.dirname(os.path.realpath(__file__)) + "/.."
     command = f"{sys.executable} -m credsweeper --path {dataset_location}/data --save-json {result_location_str} -j {j}"
-    if use_ml:
-        command += " --ml_validation"
+    if not use_ml:
+        command += " --ml_threshold skip"
```
babenek, line 11: IMHO, 'skip' is extra. 0 is enough. Or empty argument.

.github/workflows/action.yml
```diff
@@ -29,7 +29,7 @@ jobs:
       uses: Samsung/CredSweeper@main
       with:
         # args - arguments to credsweeper tool. See default values in action.yml
-        args: --path ./tests/samples/ --save-json --ml_validation
+        args: --path ./tests/samples/ --save-json
```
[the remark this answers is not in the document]
csh519: In my think removing `--ml_validation` option is failed on this PR and after merge to master it will be change with success.
Isn't it?

credsweeper/app.py
csh519: Yes, I'll handle that feature after this PR. We should change `setup.py`, but it is not that easy feature.
Actually now we support `pip install credsweepr[ml]` option for ml compatibility. So we should discuss about it. Remove `[ml]` installation option or some other way.

experiment/src/prepare_data.py
csh519: Okay. I removed `skip` attribute from `ThresholdPreset` with [4960952](https://github.com/Samsung/CredSweeper/pull/146/commits/496095207723d5e087385e74cdc890bba8c5fefc).

credsweeper/app.py
```diff
@@ -76,10 +73,15 @@ def __init__(self,
...
             from credsweeper.ml_model import MlValidator
             self.ml_validator = MlValidator(threshold=self.ml_threshold)
 
+    def _use_ml_validation(self) -> bool:
+        if self.ml_threshold != 0:
```
babenek, line 81:
```suggestion
        if 0 < self.ml_threshold:
```
btw. negative values might be passed

credsweeper/app.py
babenek: OK.
IMHO CredSweeper is used due ML

csh519: > OK. IMHO CredSweeper is used due ML

+

PR #148, opened 2022-06-28
Remove [ml] installation option

csh519:
After #146 merged. The `[ml]` installation option needs to be removed.

**Changes**
- Removed `[ml]` installation option.
- Change docs.


babenek (approved):
Ok. 
Now all ML libraries are installed as default.

kmnls (approved):
Agree

PR #149, opened 2022-07-04
Add ValueAsciiCheck filter

csh519:
There are many FP cases when scanning binary files.
So I added `ValueAsciiCheck` filter to decrease that FPs. It checks the line has any ascii character. If the line has any ascii character, it returns `False`. (`True` means filtered.)
BTW ML validation filter almost of binary cases but when user want to use CredSweeper without ML validation, `ValueAsciiCheck` filter will help them to decrease FPs.

Changes
* Add `ValueAsciiCheck` filter
* Add `GroupType.Empty` attribute


babenek (changes requested):
Add unit tests please

csh519:
> Add unit tests please

@babenek 
Actually I was working on it. :smile:
I added test with b8180b2.
Thank you for review.

csh519:
Close the PR due to further review.

PR #152, opened 2022-07-07
Add installation exception case

csh519:
There is exception case when user install `CredSweeper` on PC that didn't installed git.
So I added exception case to `install.rst` file.


babenek (approved):
OK. 
CredSweeper might scan files without git.

kmnls (approved):

PR #153, opened 2022-07-08
DataContentProvider is used to scan zip containers

babenek:
- max_depth performs scan zip archives with recursion.
- the feature is not  supported in diff due 'whatthepath' library has the limitation https://github.com/cscorley/whatthepatch/issues/4
- added two "zip-bombs" and diff-sample with binary encoded key with \0 et end


credsweeper/app.py
```diff
@@ -55,6 +61,7 @@ def __init__(self,
             ml_batch_size: int value, size of the batch for model inference
             ml_threshold: float or string value to specify threshold for the ml model
             find_by_ext: boolean - files will be reported by extension
+            max_depth: int - how deep container files will be researched
```
csh519, line 64: How about change the word? Research seems too much in this case.

```suggestion
            max_depth: int - how deep container files will be scanned
```

credsweeper/__main__.py
```diff
@@ -59,6 +59,13 @@ def get_arguments() -> Namespace:
                         help="find files by predefined extension.",
                         dest="find_by_ext",
                         action="store_true")
+    parser.add_argument("--max_depth",
+                        help="recursive search in files which are zip archives.",
+                        type=positive_int,
+                        dest="max_depth",
+                        default=0,
+                        required=False,
+                        metavar="POSITIVE_INT")
```
csh519, line 68: `max_depth` naming is good, but in my think it just used on zip files, so what about `zip_scan_depth` or `zip_depth` or so on..?
I think `max_depth` can be read scanning file depth..

credsweeper/app.py
```diff
@@ -217,6 +225,14 @@ def file_scan(self, content_provider: ContentProvider) -> List[Candidate]:
             # Skip the file scanning and create fake candidate because the extension is suspicious
             candidates.append(Candidate.get_dummy_candidate(self.config, content_provider.file_path))
 
+        elif self.config.max_depth > 0 and isinstance(content_provider, TextContentProvider):
+            # Feature to scan files which might be containers
+            data = Util.read_data(content_provider.file_path)
+            if data:
+                data_provider = DataContentProvider(data=data, file_path=content_provider.file_path)
+                # use limit with 1Gb
+                candidates = self.data_scan(data_provider, self.config.max_depth, 1 << 30)
```
csh519, line 234: Set limitation to 1Gb with `1 << 30` is good. But what about use `Constant` class?
I think it would be better adding `ZIP_SCAN_LIMITATION` attribute and using it.

credsweeper/app.py
```diff
@@ -227,6 +243,63 @@ def file_scan(self, content_provider: ContentProvider) -> List[Candidate]:
...
+                data_provider: DataContentProvider object may be a container
+                depth: maximal level of recursion
+                recursive_limit: maximal bytes of opened files to prevent recursive zip-bomb attack
+        """
+        candidates: List[Candidate] = []
+        logging.debug(f"Start scan data: {data_provider.file_path} {len(data_provider.data)} bytes")
+
+        if 0 > depth:
+            # break recursion if maximal depth is reached
+            logging.error(f"bottom reached {data_provider.file_path} recursive_limit:{recursive_limit}")
```
csh519, line 259: Is it good to logging as error..? I think there are many cases that over `recursive_limit`..
So how about set logging level as a debug?
```suggestion
            logging.debug(f"bottom reached {data_provider.file_path} recursive_limit:{recursive_limit}")
```

csh519 (changes requested):
I reviewed some but I stopped reviewing the PR..
Because I couldn't understand why the `data_scan()` is needed.
I think `file_scan()` can handle this feature and also `content_provider` that in app.py#187 can handle `DataContentProvider`.
So I think this PR increase complexity of the code. Could you please check this again?

tests/test_app.py
```diff
@@ -243,3 +244,44 @@ def test_find_by_ext_n(self) -> None:
...
+    def test_patch_multifile_p(self) -> None:
+        dir_path = os.path.dirname(os.path.realpath(__file__))
+        target_path = os.path.join(dir_path, "samples", "multifile.patch")
+        proc = subprocess.Popen([sys.executable, "-m", "credsweeper", "--diff_path", target_path, "--log", "silence"],
+                                stdout=subprocess.PIPE,
+                                stderr=subprocess.PIPE)
+        _stdout, _stderr = proc.communicate()
+        output = " ".join(_stdout.decode("UTF-8").split())
+        print(output)
+        assert output == 'returning from dumb path returning from dumb path returning from dumb path returning from dumb path'
```
csh519, line 286: I checked this mysterious string.. and find out it comes from `whatthepatch`. (whatthepatch>patch.py#L248)
I think it is not good to add stdout TC that comes from other library.. not ours..
So how about remove those assert case?

credsweeper/__main__.py
babenek: I suppose, the argument might be used for scan others containers like ISO images, tar-archives and so on.
Probably ```containers_depth``` or just ```depth```
Otherwise in future we have to change ```zip_``` prefix to avoid confuses.

tests/test_app.py
babenek: I'll move the test of ```whatthepath``` lib to separated test for the case - ok?

csh519: Okay that's good idea.

credsweeper/__main__.py
csh519: Checked `max_depth` has been changed to `depth` and resolve conversation.

credsweeper/app.py
csh519: `RECURSIVE_SCAN_LIMITATION` has been added and resolve conversation.

credsweeper/app.py
```diff
@@ -227,6 +243,63 @@ def file_scan(self, content_provider: ContentProvider) -> List[Candidate]:
...
+            return candidates
+
+        depth -= 1
+
+        if self.config.find_by_ext and FilePathExtractor.is_find_by_ext_file(self.config, data_provider.file_path):
+            # Skip scanning file and makes fake candidate due the extension is suspicious
+            candidates.append(Candidate.get_dummy_candidate(self.config, data_provider.file_path))
```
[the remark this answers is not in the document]
csh519: Totally right. I missed `data_scan()` working with recursively.

credsweeper/app.py
```diff
@@ -227,6 +243,63 @@ def file_scan(self, content_provider: ContentProvider) -> List[Candidate]:
...
+            try:
+                with zipfile.ZipFile(io.BytesIO(data_provider.data)) as zf:
+                    for zfl in zf.infolist():
+                        file_path = f"{data_provider.file_path}/{zfl.filename}"
+                        # skip directory
+                        if "/" == file_path[-1:]:
+                            continue
+                        if FilePathExtractor.check_exclude_file(self.config, file_path):
+                            continue
+                        if 0 > recursive_limit - zfl.file_size:
```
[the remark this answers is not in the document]
csh519: `recursive_limit_size` accepted. Resolve conversation.

credsweeper/file_handler/data_content_provider.py
```diff
@@ -0,0 +1,32 @@
...
+    """
+
+    def __init__(self, data: bytes, file_path: Optional[str] = None) -> None:
+        super().__init__(file_path if file_path is not None else "")
+        self.__data = data
+
+    @property
+    def data(self) -> bytes:
+        """data getter"""
+        return self.__data
```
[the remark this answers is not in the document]
csh519: The code block accepted. Resolve conversation.

tests/test_app.py
```diff
@@ -243,3 +244,44 @@ def test_find_by_ext_n(self) -> None:
...
+
+    def test_patch_multifile_p(self) -> None:
+        dir_path = os.path.dirname(os.path.realpath(__file__))
+        target_path = os.path.join(dir_path, "samples", "multifile.patch")
+        proc = subprocess.Popen([sys.executable, "-m", "credsweeper", "--diff_path", target_path, "--log", "silence"],
+                                stdout=subprocess.PIPE,
+                                stderr=subprocess.PIPE)
+        _stdout, _stderr = proc.communicate()
+        output = " ".join(_stdout.decode("UTF-8").split())
+        print(output)
```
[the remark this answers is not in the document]
csh519: The method has been removed and resolve conversation.

csh519 (approved):
Thank you for the quick fix and good feature(container scanning).

LGTM :+1:

credsweeper/__main__.py
```diff
@@ -168,7 +176,7 @@ def main() -> None:
         logging.info(f"Run analyzer on path: {args.path}")
         content_provider: FilesProvider = TextProvider(args.path, skip_ignored=args.skip_ignored)
         scan(args, content_provider, args.json_filename)
-    if args.diff_path:
+    elif args.diff_path:
```
kmnls, line 179: This change should be in the separate PR, I suppose

babenek: Just to organize code. Arguments must be resolved in parcer
```
    group = parser.add_mutually_exclusive_group(required=True)
    group.add_argument("--path", nargs="+", help="file or directory to scan", dest="path", metavar="PATH")
    group.add_argument("--diff_path", nargs="+", help="git diff file to scan", dest="diff_path", metavar="PATH")
```

credsweeper/utils/util.py
```diff
@@ -218,3 +219,41 @@ def preprocess_file_diff(changes: List[DiffDict]) -> List[DiffRowData]:
...
+    def is_zip(data: bytes) -> bool:
+        """According https://en.wikipedia.org/wiki/List_of_file_signatures"""
+        if isinstance(data, bytes) and 3 < len(data):
+            # PK
+            if 0x50 == data[0] and 0x4B == data[1]:
+                if 0x03 == data[2] and 0x04 == data[3]:
+                    return True
+                # empty archive - no sense to scan
+                elif 0x05 == data[2] and 0x06 == data[3]:
+                    return True
```
kmnls, line 233: No need to scan: we may safely skip it from scanning procedure by returning False
```suggestion
                    return False
```

babenek: It skips scanning the file with TextContentProvider.

PR #154, opened 2022-07-08
Refactoring and unit tests

babenek:
Some refactoring and tests to reduce next PR ( #153 ) complexity

### DIFF
--- a/credsweeper/app.py
+++ b/credsweeper/app.py
@@ -214,7 +214,7 @@ class CredSweeper:
         logging.debug(f"Start scan file: {content_provider.file_path}")
 
         if self.config.find_by_ext and FilePathExtractor.is_find_by_ext_file(self.config, content_provider.file_path):
-            # Skip scanning file and makes fake candidate due the extension is suspicious
+            # Skip the file scanning and create fake candidate because the extension is suspicious
             candidates.append(Candidate.get_dummy_candidate(self.config, content_provider.file_path))
 
         else:
--- a/credsweeper/app.py
+++ b/credsweeper/app.py
@@ -5,11 +5,10 @@ import multiprocessing
 import os
 import signal
 import sys
-import regex
 
-from credsweeper.common.constants import KeyValidationOption, ThresholdPreset, DEFAULT_ENCODING, Severity
+from credsweeper.common.constants import KeyValidationOption, ThresholdPreset, DEFAULT_ENCODING
 from credsweeper.config import Config
-from credsweeper.credentials import Candidate, CredentialManager, LineData
+from credsweeper.credentials import Candidate, CredentialManager
 from credsweeper.file_handler.content_provider import ContentProvider
 from credsweeper.file_handler.file_path_extractor import FilePathExtractor
 from credsweeper.file_handler.files_provider import FilesProvider
@@ -78,14 +77,20 @@ class CredSweeper:
         self.ml_threshold = ml_threshold
         self.ml_validator = None
 
+    # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # #
+
     def _use_ml_validation(self) -> bool:
         if isinstance(self.ml_threshold, float) and self.ml_threshold <= 0:
             return False
         return True
 
+    # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # #
+
     # the import cannot be done on top due
     # TypeError: cannot pickle 'onnxruntime.capi.onnxruntime_pybind11_state.InferenceSession' object
     from credsweeper.ml_model import MlValidator
+
+    # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # #
 
     @property
     def ml_validator(self) -> MlValidator:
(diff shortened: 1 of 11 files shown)


credsweeper/app.py
```diff
@@ -78,15 +77,21 @@ def __init__(self,
         self.ml_threshold = ml_threshold
         self.ml_validator = None
 
+    # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # #
```
csh519, line 80: Why those lines are added? It just for readability..? :thinking:

tests/ml_model/test_features.py
```diff
@@ -5,28 +5,31 @@
...
     test_entropy = RenyiEntropy('hex', 0, norm=True)
-    probabilities = test_entropy.get_probabilities('Quick brown fox jumps over the lazy dog')
+    probabilities = test_entropy.get_probabilities(AZ_STRING)
     print(probabilities)
     assert len(probabilities) == 6
-    expected_max = [0.14285715, 0.14285715, 0.14285715, 0.14285715, 0.28571430, 0.14285715]
-    expected_min = [0.14285714, 0.14285713, 0.14285713, 0.14285713, 0.28571428, 0.14285713]
+    expected_max = [0.12500001, 0.12500001, 0.12500001, 0.12500001, 0.37500001, 0.12500001]
+    expected_min = [0.12499999, 0.12499999, 0.12499999, 0.12499999, 0.37499999, 0.12499999]
     for n in range(6):
```
csh519, line 18: Could you explain why this `expected_min/max` values are changed?

tests/test_app.py
```diff
@@ -192,7 +194,7 @@ def test_find_by_ext_p(self) -> None:
             for f in [".pem", ".crt", ".cer", ".csr", ".deR"]:
                 file_path = os.path.join(tmp_dir, f"dummy{f}")
                 assert not os.path.exists(file_path)
-                open(file_path, "w").write("The quick brown fox jumps over the lazy dog")
+                open(file_path, "wb").write(AZ_DATA)
```
csh519, line 197: Is there any differences between write mode and binary write mode? Please let me know about this..

credsweeper/app.py
babenek: Yes, it helps review large files + some tools like 'meld' use the delimiters like anchors and recognize diff more correctly.

tests/__init__.py
```diff
@@ -6,3 +6,7 @@
 
 # credentials count after post-processing
 SAMPLES_POST_CRED_COUNT: int = 18
+
+# well known string with all latin letters
+AZ_DATA = b"The quick brown fox jumps over the lazy dog"
+AZ_STRING = AZ_DATA.decode(encoding="ascii")
```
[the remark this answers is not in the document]
babenek: TEST_DATA is more common name. Probably it might be used in specific test cases.
The pangram (a-z letters) is well known and their hash might be found in many references.
When the bytes are written then hash of  file must be same like in the references.  
Sometime with text working may be mistakes like:
C/C++ - \0 is written in tail
bash - \n appended in echo case ...

tests/ml_model/test_features.py
babenek: Previous text was without "The " - so 'h' was skipped.

tests/test_app.py
babenek: No. There is no difference. The test requires only not-empty files to find them by extension. Possibly, bytes are not required extra conversion.

credsweeper/app.py
```diff
@@ -177,42 +198,34 @@ def __multi_jobs_scan(self, content_providers: Union[List[DiffContentProvider],
...
-            scan_context = file_provider.get_analysis_target()
-            return self.scanner.scan(scan_context)
-        except UnicodeDecodeError:
-            logging.warning(f"Can't read file content from \"{file_provider.file_path}\".")
-            return []
+        candidates: List[Candidate] = []
+        logging.debug(f"Start scan file: {content_provider.file_path}")
+
+        if self.config.find_by_ext and FilePathExtractor.is_find_by_ext_file(self.config, content_provider.file_path):
+            # Skip scanning file and makes fake candidate due the extension is suspicious
```
kmnls, line 217:
```suggestion
            # Skip the file scanning and create fake candidate because the extension is suspicious
```

kmnls (dismissed):

tests/ml_model/test_features.py
csh519: Oh.. I missed "The".

tests/test_app.py
csh519: If there is no difference, what about change "wb" option to "w"?
Almost all file write options that in CredSweeper are using "w".

credsweeper/app.py
csh519: Okay.. I agree.

tests/test_app.py
babenek: Ok. This suite was rollback to use 'w'.

csh519: > Ok. This suite was rollback to use 'w'.

@babenek 
It seems there is no change regarding 'w' option..
Please check.

csh519 (approved):
Thank you for refactoring!

LGTM :+1:

credsweeper/app.py
```diff
@@ -177,42 +198,34 @@ def __multi_jobs_scan(self, content_providers: Union[List[DiffContentProvider],
...
-                    patterns=[regex.compile(".*")],  #
-                    rule_name="Dummy candidate",  #
-                    severity=Severity.INFO,  #
-                    config=self.config)
-                return [candidate]
-
-        try:
-            scan_context = file_provider.get_analysis_target()
-            return self.scanner.scan(scan_context)
-        except UnicodeDecodeError:
```
kmnls, line 213: You have removed exception check here.
Is this done by purpose?

babenek: it is done inside get_analysis_target and  scan. Moreover, the line was not covered with fuzzing that generated many various bytes that may throw the exception. 
So, the exception is extra here.

kmnls (approved):

iuriimet (approved):
Agree

PR #156, opened 2022-07-18
Add --save-xlsx option

csh519:
Add save xlsx option to CredSweeper.
You can run save xlsx option by adding `--save-xlsx` on command line.
- Added `--save-xlsx` option
- Fix wrong import ordering
- Fix wrong variable using

Below packages are newly included.
- pandas
- openpyxl

@babenek Thanks to solve the pytest mock issue 👍


babenek (approved):
Looks good. XLSX report is created.

mykolaf (approved):
:+1:

credsweeper/__main__.py
```diff
@@ -106,6 +106,12 @@ def get_arguments() -> Namespace:
                         const="output.json",
                         dest="json_filename",
                         metavar="PATH")
+    parser.add_argument("--save-xlsx",
```
kmnls, line 109: In the report generation code either json or xml is generated.
In the argument parsing there is no info about the mutual exclusivity of these options, isn't it?

csh519: I fixed it by 7b231b5.
Please check.

kmnls (approved):

csh519:
> Probably, there should be an error message if both are specified, shouldn't be?

@kmnls 
I tested with below command, and it worked fine :)

``` bash
~$ python -m credsweeper --path ~/test/ --save-json json.json --save-xlsx xlsx.xlsx
```

PR #157, opened 2022-07-19
Multiline credential case

babenek:
Many sources types long lines may be split to multiple with \ symbol
The PR may fix some false negative cases.


tests/test_app.py
```diff
@@ -275,3 +275,21 @@ def test_zip_p(self) -> None:
...
+            json_filename = os.path.join(tmp_dir, "report.json")
+            proc = subprocess.Popen([
+                sys.executable, "-m", "credsweeper", "--log", "debug", "--path", sample_path, "--save-json",
+                json_filename, "--ml_threshold", "0"
+            ],
+                                    stdout=subprocess.PIPE,
+                                    stderr=subprocess.PIPE)
+            _stdout, _stderr = proc.communicate()
+            print(_stdout)
+            print(_stderr)
```
csh519, line 291: It seems those lines can be remove.
Please check.

tests/test_app.py
```diff
@@ -275,3 +275,21 @@ def test_zip_p(self) -> None:
             with open(json_filename, "r") as json_file:
                 report = json.load(json_file)
                 assert len(report) == SAMPLES_POST_CRED_COUNT + 1
+
+    def test_dup_p(self) -> None:
+        with tempfile.TemporaryDirectory() as tmp_dir:
+            sample_path = os.path.join(os.path.dirname(os.path.realpath(__file__)), "samples", "test_dup")
+            json_filename = os.path.join(tmp_dir, "report.json")
+            proc = subprocess.Popen([
+                sys.executable, "-m", "credsweeper", "--log", "debug", "--path", sample_path, "--save-json",
```
csh519, line 284: `--log` option looks like used for debugging. Could you please remove `--log` option?

credsweeper/file_handler/content_provider.py
```diff
@@ -28,7 +28,23 @@ def file_path(self) -> str:
...
+                else:
+                    previous_target = AnalysisTarget(new_line, i + 1, lines, self.file_path)
+            else:
+                if previous_target:
+                    previous_target.line += line
+                    targets.append(previous_target)
+                    previous_target = None
+            # Anyway add original line to analysis
             target = AnalysisTarget(line, i + 1, lines, self.file_path)
             targets.append(target)
```
csh519, line 47: I'm worried about adding original line to `targets` at every iteration..
If there is some code block like below, then user will receive duplicate report every time.
I think it would be good to add only merged line(`previous_target` + `line`) and original line(if there is no `'\\'`).
How do you think?

**Code block**

``` python
aws_client_id = \
"AKIAIANNKJJHPERXR0LA"
# This is example key
```

**Report**
``` bash
rule: AWS Client ID / severity: high / line_data_list: [line: 'aws_client_id = "AKIAIANNKJJHPERXR0LA"  ' / line_num: 1 / path: xx / value: 'AKIAIANNKJJHPERXR0LA' / entropy_validation: False] / api_validation: NOT_AVAILABLE / ml_validation: VALIDATED_KEY
rule: AWS Client ID / severity: high / line_data_list: [line: '"AKIAIANNKJJHPERXR0LA"  ' / line_num: 2 / path: xx / value: 'AKIAIANNKJJHPERXR0LA' / entropy_validation: False] / api_validation: NOT_AVAILABLE / ml_validation: VALIDATED_KEY
```

babenek: I did not replace target due we have limit for string ````MAX_LINE_LENGTH = 1500``` - if a string is oversize - it is skipping :( even without warning.
So, merged string may be too long and possibly credentials are skipped.
Thus it is not a good source code, but credentials are badly. 
My opinion is to keep possibly duplicates when merged line is too long + to show the warning.

csh519: Hmm.. I thought it more.. and still I think it is not good to keeping duplicates.. (Other services that using CredSweeper will be affected.)

So I suggest like this..

1. Manage and gather previous lines with list.
2. If gathered line exceed `MAX_LINE_LENGTH = 1500`.
   - True: Append each gathered lines to `targets`.
   - Else: Append gathered line to `targets`.

It might be a solution for keeping long line and each lines.

PR #162, opened 2022-07-25
Fix and test AWS multi pattern

babenek:
Fixed false-positive case in huge base64 encoded lines after AWS credential
According [User Guide for Version 2](https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-welcome.html) there only 40 characters must be provided or more if [token](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-files.html)


credsweeper/rules/config.yaml
```diff
@@ -26,7 +26,7 @@
   type: pattern
   values:
   - (?P<value>(AKIA|ASIA)[0-9A-Z]{16})
-  - (?P<value>[0-9a-zA-Z/+]{40})
+  - ([^0-9a-zA-Z/+])(?P<value>[0-9a-zA-Z/+]{40})([^0-9a-zA-Z/+])
```
csh519, line 29: Could you explain why `([^0-9a-zA-Z/+])` strings are added before and after original pattern?
If AWS only use 40 characters to secret, I think original one is better than this.

babenek: Unfortunately - {40} means that the value contains 40 symbols, but left and right symbols are not involved. They might be various.
As you can see from test sample - it is.
Probably we shall test others patterns in the PR to reduce FP.

csh519: Okay..
Could you test them by CredData?
There is benchmark function.
Changing rule config needs to prove improvement and not to increase FN.

credsweeper/rules/config.yaml
```diff
@@ -25,8 +25,8 @@
   severity: high
   type: pattern
   values:
-  - (?P<value>(AKIA|ASIA)[0-9A-Z]{16})
-  - (?P<value>[0-9a-zA-Z/+]{40})
+  - ((?:[^0-9A-Za-z])|$)(?P<value>(AKIA|ASIA)[0-9A-Z]{16})((?:[^0-9A-Za-z])|$)
+  - ((?:[^0-9A-Za-z/+])|$)(?P<value>[0-9a-zA-Z/+]{40})((?:[^0-9A-Za-z/+])|$)
```
babenek, line 29:
```suggestion
  - (^|(?:[^0-9A-Za-z]))(?P<value>(AKIA|ASIA)[0-9A-Z]{16})((?:[^0-9A-Za-z])|$)
  - (^|(?:[^0-9A-Za-z/+]))(?P<value>[0-9a-zA-Z/+]{40})((?:[^0-9A-Za-z/+])|$)
```

credsweeper/rules/config.yaml
babenek: After update here is changes: https://github.com/babenek/CredSweeper/commit/762e6367a53109af4d13e649d449fd950b8a158d
Moreover, i compared results and found - many 'AWS ID' were found in lines which are longer for 1 symbol (e.g. AKIA0123456789ABCDEFG)

csh519 (commented):
It seems this PR decrease benchmark performance..
How do you think?

babenek:
Well, possibly AWS multi may be not limited it 40 symbols because it may be 162 symbols for [token](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-files.html)

babenek:
Final variant. It removed real one FP from CredData. Benchmark results (since release) https://github.com/babenek/CredSweeper/commit/87b4e3e992810c3b150bc809d5af937e416b11a1

iuriimet (approved):
LGTM

PR #165, opened 2022-07-29
Update action.yml

babenek:
Using 'main' action will found 18 items. Like in tests 
SAMPLES_POST_CRED_COUNT: int = 18


iuriimet (approved):
Agree

kmnls (approved):

PR #168, opened 2022-08-02
Catch more zipfile exceptions.

babenek:
- zipfile might throw more exceptions than documented - catch all
- Consumer size was extend to 2048 to generate string with size more than 1500 chars.
- corpus dir was moved inside fuzz/
- main fuzzed function is data_scan. it produces more branches and found uncaught exception like https://github.com/python/cpython/issues/71351

old:TOTAL                                                         2271    462    795     70    77%
new:TOTAL                                                         2271    416    795     71    80%


csh519 (approved):
Fuzzing code is getting better.

LGTM :+1:

iuriimet (dismissed):
Looks good

kmnls (changes requested):
content_provider is out of PR scope

csh519 (approved):
Now the code will catch all `zipfile` exception.

LGTM :+1:

kmnls (approved):

Issue #69, opened 2022-01-26
Add minimal severity, additional rules, log config, and config arguments


meanrin:
Some files in the scanner is modifiable in theory, but is a pain to eddit if package installed using pip

This include: 
- `credsweeper/secret/config.json` - file that lists exclude pats, and what fields should be saved in the final report
- `credsweeper/secret/log.yaml` - file specifies log format, max size of log file if any is used, log file location (none by default), etc

So I propose to add something similar to:
```
--config [PATH]          path of config file (default: credsweeper/secret/config.json)
--log-config [PATH]      path of log config file (default: credsweeper/secret/log.yaml)
```

------------------------

In addition this 2 flags might be useful:

`minimal-severity` can be a useful flag for people who do not want to edit config, but just don't want to see info or low severity alerts

```
--minimal-severity            Hides any report with severity lower than specified. Any of ['info', 'low', 'medium', 'high', 'critical']. Default: info
```

`additional-rules` might allow to add new rules in addition to existing one, without copying base file

```
--additional-rules [PATH]     Add new rules to the one already listed in `credsweeper/rules/config.yaml` without overriding the old one 
```

That would allow us to avoid next scenario:
1. User copied our rule file and manually adds a new rule
2. We edit main rule file (for example fixing old regex or adding new rules) in a next release
3. Now user cannot use our newly edited main rule file, until they manually update their own edited copy 

By adding `additional-rules` that do not override the main one this scenario can be avoided

babenek:
Enhance proposal:
1. put all default values into code when we not use package files. It speedup fuzzing a bit and improve testing.
2. migrate from YAML to JSON
3. add --export_... (log, config, rules) to make a template for user

PR #170, opened 2022-08-03
Config

babenek:


csh519:
Sorry for commenting on draft PR..

I wonder why you make `config.json` to `config.py`..
Actually i think it is not that improve CredSweeper.

babenek:
The approach speed up unit test for ~43%, but only 3%(320s -> 310s) for multiple --jobs for CredData collection. 
Speedup 6% is for multiple launch for each file. e.g.
```
s=$(date +%s.%N); for f in $(ls tests/samples);do credsweeper --path $f; done; echo "$(date +%s.%N) - $s" | bc -l | sed 's/^\./0./g'
```
Only using **single** job for CredData collection slowdown less than 1%. 

Benefits:
1. Remove external PyYAML  module
2. Configs unification. Currently they are:
- json (almost native and easy migration to python's dict and back)
- Yaml (one benefit - simple escaping backslash in regex. may be used as alternative)
- TXT (keywords ? - i move this into Config class)
3. Feature for export a default config and load. Currently only rules may be load but getting original rules is a quest for user.
4. Less files are in package. May be ML model including in sources is extra...
5. Plan to move all variables from Credsweeper.__init__ arguments to Config to support load_config feature. It helps avoid fixture using in tests and manage the tests clearly. 


Anyway, the PR is only draft with huge changes and I'll make smallest PRs. 
The thread may be used for discussion

csh519:
@babenek 

Thank you for explanation.

All benefits are reasonable and I agreed.
But I worried about user experience.. Maybe users would prefer `config.yaml` to `config.py`..

babenek:
```config.py``` is used for 'built-in' config and for exporting. The config may be exported to json and modified by user. Then, user will be able to use the file as custom config. YAML may be converted to json and back very easy.

csh519:
@babenek 

Okay. I agreed.

meanrin (changes requested):
My issues with this PR:

1. It's almost impossible to review it. Please be reasonable and don't:
- make a PR without ANY description
- make a PR with 8k lines changed 

2. You moved model binary files and model config into the code
This change is akin moving `.jpg` or `.dll` files to inside the python code 
It’s just much more complicated and obscure 
See full response at https://github.com/Samsung/CredSweeper/pull/187#pullrequestreview-1069606102 

3. I don't find unit test speed up as a good motivation for such huge architectural change
- can you provide what percent of time spent on config reading during a real project scan? 
- config files provide simplicity, readability, do not interfere with code coverage and they are easy to edit
- it you concerned about speed and you identified config reading as a bottleneck, please conciser what other possible solutions might be, before rewriting 8k lines of code

4. Moving config to code is a literal anti-pattern:
> Hard coding is the poor practice where we store configuration or input data in the source code rather than obtaining it from a configuration file or another external source.
https://medium.com/codex/software-anti-patterns-e5212d3dd421

meanrin:
In general
@babenek please discuss possible solutions with other maintainers before changing more than 8k lines in 34 files in a single request

babenek:
@meanrin i created the PR as draft and gave as reference only. 
Yes, we've discussed. 
Binary file including was rollback.
We are trying organize issue when configuration parameters are scattered in code.
e.g. hardcoded NOT_ALLOWED_PATH, a Zoo (txt, json,yaml). logging has another native format...
Possibly we will move some variables (e.g. pool_count AKA jobjs) to config but arguments should have highest priority.
This is a solution of discuss in #183
When user wants to scan a repo with --depth option - we have to remove .zip extension from exclude list. 
May be this is not what user wants, may be not. 
So, proposal is: export default config to user path, user makes changes and runs  app with custom config.
Between two formats json (standard library) or yaml(external package, only regex is a bit easiest) i choose native python dictionary as compromise. It might be transform to json or yaml as user wants. 
But extra reading from file for default is extra. IMHO. 
The hardcoded way is not for freeze due import of custom will be available.
Speedup was 5% for CredData and 45% for tests.
Coverage of default dict is only 1 line - look to the artifact https://github.com/Samsung/CredSweeper/suites/7775384179/artifacts/327166145.

meanrin:
> Yes, we've discussed.

My bad
Please provide a reference to the discussion so everyone can check it

babenek:
> 
1. with @kmnls, @iuriimet  voice
2. 7days above in PR

meanrin:
> 

@babenek may we see it in any documented form?
And why on earth only 3 maintainers participated in a discussion that rewrites architecture, remove all config files and changes up to 8k lines of code? Are there no other stakeholders, maintainers, SQA people involved in the project?

ARKAD97 (changes requested):
Agreed with @meanrin on all points.
Config files should never be moved into the python code

babenek:
@meanrin #69 i proposed the changes tree weeks ago

PR #177, opened 2022-08-08
Add completion phrase

csh519:
There was no completion phrase, and some users wondered the scanning finished well.

So I added completion phrase at the end of scanning process.

### DIFF
--- a/credsweeper/__main__.py
+++ b/credsweeper/__main__.py
@@ -178,6 +178,7 @@ def scan(args: Namespace, content_provider: FilesProvider, json_filename: Option
 
 def main() -> None:
     """Main function"""
+    start_time = time.time()
     args = get_arguments()
     os.environ["LOG_LEVEL"] = args.log
     Logger.init_logging(args.log)
@@ -200,12 +201,12 @@ def main() -> None:
     else:
         logging.error("Not specified 'path' or 'diff_path'")
 
-    for k, v in result.items():
-        print(f"{k}: {v}")
+    if len(result):
+        for k, v in result.items():
+            print(f"{k}: {v}")
+        end_time = time.time()
+        print(f"Time Elapsed: {end_time - start_time}s")
 
 
 if __name__ == "__main__":
-    start_time = time.time()
     main()
-    end_time = time.time()
-    print(f"Time Elapsed: {end_time - start_time}s")
--- a/credsweeper/__main__.py
+++ b/credsweeper/__main__.py
@@ -1,4 +1,5 @@
 import os
+import time
 from argparse import ArgumentParser, ArgumentTypeError, Namespace
 from typing import Any, Union, Optional
 
@@ -149,7 +150,7 @@ def get_json_filenames(json_filename: str):
 
 
 def scan(args: Namespace, content_provider: FilesProvider, json_filename: Optional[str],
-         xlsx_filename: Optional[str]) -> None:
+         xlsx_filename: Optional[str]) -> int:
     """Scan content_provider data, print results or save them to json_filename is not None
 
     Args:
@@ -172,7 +173,7 @@ def scan(args: Namespace, content_provider: FilesProvider, json_filename: Option
                               find_by_ext=args.find_by_ext,
                               depth=args.depth,
                               size_limit=args.size_limit)
-    credsweeper.run(content_provider=content_provider)
+    return credsweeper.run(content_provider=content_provider)
 
 
 def main() -> None:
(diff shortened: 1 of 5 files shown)


tests/test_app.py
```diff
@@ -88,6 +90,8 @@ def test_it_works_with_patch_p(self) -> None:
                         / entropy_validation: False]
                     / api_validation: NOT_AVAILABLE
                     / ml_validation: VALIDATED_KEY\n
+                    Scanning finished / Detected Credentials : 1\n
+                    Scanning finished / Detected Credentials : 0
```
babenek, line 94: Possibly the double lines are confusing too.
Proposal: add argument to the function for printing result.
e.g. "Added diff scan finished" , "Removed diff..." Path scan finished"

babenek (changes requested):
Duplicate message "Scanning finished" is still confusing.
Probably we have a chance to return exit code OK when no credential found and ERROR - when it is.
please, check https://github.com/Samsung/CredSweeper/pull/175

tests/test_app.py
csh519: I'll remove `Scanning finished` phrase and just left `Detected Credentials` for each case(Diff scan, Normal scan).

credsweeper/app.py
```diff
@@ -356,3 +356,5 @@ def export_results(self) -> None:
         if is_exported is False:
             for credential in self.credential_manager.get_credentials():
                 print(credential)
+
+        print(f"Scanning finished / Detected Credentials : {len(self.credential_manager.get_credentials())}")
```
kmnls, line 360: Probably the logging API should be used, shan't be?

tests/test_app.py
meanrin: Alternatively:
Change `def scan` in `__main__.py` to return number of creds by checking something like `return len(credsweeper.credential_manager.get_credentials())`
https://github.com/Samsung/CredSweeper/blob/0d8ae1e8f112be0a5cbea2ab3f1037c90f2934ae/credsweeper/__main__.py#L151 

Then print can be done in `def main`, without passing any additional arguments:
https://github.com/Samsung/CredSweeper/blob/0d8ae1e8f112be0a5cbea2ab3f1037c90f2934ae/credsweeper/__main__.py#L187
https://github.com/Samsung/CredSweeper/blob/0d8ae1e8f112be0a5cbea2ab3f1037c90f2934ae/credsweeper/__main__.py#L193
https://github.com/Samsung/CredSweeper/blob/0d8ae1e8f112be0a5cbea2ab3f1037c90f2934ae/credsweeper/__main__.py#L197

Bonus: 
`def main` can also be used o measure execution time, and so exec time can also be added to the final output

kmnls (dismissed):

tests/test_app.py
csh519: @meanrin 

Good suggestion.
I'll refer to it when I modify code.
Thank you.

babenek: @meanrin linux may shrink the number.

babenek: $ python -c "import sys; sys.exit(1223)"
$ echo $?
199

meanrin: > @meanrin linux may shrink the number.

Nah, I was not talking about exit code. Just a potential location to check the creds 


> $ python -c "import sys; sys.exit(1223)" $ echo $? 199

I think linux using only 1 byte for the exit code, so it just overflows anything greater than 255 

```
$ python -c "import sys; sys.exit(255)"
$ echo $?
255

$ python -c "import sys; sys.exit(256)"
$ echo $?
0

$ python -c "import sys; sys.exit(257)"
$ echo $?
1
```

babenek: @meanrin unfortunately we have to support windows and mac. https://docs.python.org/3.4/library/os.html#os.EX_OK - Unix only. 
So, I think only 0 and 1 are valuable for exit codes and i propose use as is.
Ok - even credentials were found. Failure exit might brake some users scripts.
Failure - if CredSweeper cannot work totally. Probably, some errors may be accumulated like file access errors.

csh519: I changed the code by [656a39a](https://github.com/Samsung/CredSweeper/pull/177/commits/656a39ac30d3e354dcbff17a55cf35bf39f45d6a).
Please check.

credsweeper/__main__.py
```diff
@@ -181,23 +182,30 @@ def main() -> None:
...
         logging.error("Not specified 'path' or 'diff_path'")
 
+    for k, v in result.items():
+        print(f"{k}: {v}")
+
 
 if __name__ == "__main__":
+    start_time = time.time()
     main()
+    end_time = time.time()
```
babenek, line 210: According #175 the lines will be skipped.

credsweeper/__main__.py
```diff
@@ -181,23 +182,30 @@ def main() -> None:
...
+    print(f"Time Elapsed: {end_time - start_time}s")
```
babenek, line 211: may be the text should be displayed if there were no errors

babenek (commented):
Please, move the chronometry  into main and display only if result exists. It is extra in error case.

credsweeper/__main__.py
csh519: I moved them into `main()`.
Please check 4c21bc0.

babenek (approved):
OK. No conflicts i suppose

tests/validation/test_square_client_id_validation.py
```diff
@@ -22,6 +22,6 @@ def line_data_list(self) -> List[LineData]:
 
     @pytest.mark.skipif(environ.get("CIRCLE_PROJECT_USERNAME") is not None,
                         reason="Server blocking requests from CI server")
-    def test_verify_n(self, line_data_list: pytest.fixture) -> None:
+    def test_verify_p(self, line_data_list: pytest.fixture) -> None:
         validation_result = SquareClientIdValidation.verify(line_data_list)
-        assert validation_result is KeyValidationOption.INVALID_KEY
+        assert validation_result is KeyValidationOption.VALIDATED_KEY
```
babenek, line 27: BTW, how about #181 ?

csh519: I think mocking is not a good solution for this tests...

I think it would be better to get a response from an actual server.

babenek: Actually, UT shall use mocking only. 
Tests with wild servers are called "integration", "functional" or something like this. Not "unit-test"

csh519: I agreed.
Then mocking would be good in this case...

kmnls (approved):

PR #104, opened 2022-04-26
Fix JSON Web Token regex to reduce FN cases

csh519:
There were not many JWT cases in perfect form.
So I fixed `JSON Web Token` regex by remove `.eyj` string set and here is result.

I tested with [CredData](https://github.com/Samsung/CredData).

Before:
``` bash
result_cnt : 3981, lost_cnt : 0, true_cnt : 3662, false_cnt : 319
credsweeper -> TP : 3662, FP : 319, TN : 19454380, FN : 921, FPR : 0.0000163971, FNR : 0.2009600698, ACC : 0.9999362772, PRC : 0.9198693796, RCL : 0.7990399302, F1 : 0.8552078468
credsweeper Private Key -> TP : 953, FP : 0, TN : 4, FN : 39, FPR : 0E-10, FNR : 0.0393145161, ACC : 0.9608433735, PRC : 1.0000000000, RCL : 0.9606854839, F1 : 0.9799485861
credsweeper Predefined Pattern -> TP : 273, FP : 2, TN : 40, FN : 54, FPR : 0.0476190476, FNR : 0.1651376147, ACC : 0.8482384824, PRC : 0.9927272727, RCL : 0.8348623853, F1 : 0.9069767442
credsweeper Generic Token -> TP : 281, FP : 5, TN : 598, FN : 52, FPR : 0.0082918740, FNR : 0.1561561562, ACC : 0.9391025641, PRC : 0.9825174825, RCL : 0.8438438438, F1 : 0.9079159935
credsweeper Generic Secret -> TP : 975, FP : 4, TN : 214, FN : 81, FPR : 0.0183486239, FNR : 0.0767045455, ACC : 0.9332810047, PRC : 0.9959141982, RCL : 0.9232954545, F1 : 0.9582309582
credsweeper Password -> TP : 986, FP : 129, TN : 4167, FN : 409, FPR : 0.0300279330, FNR : 0.2931899642, ACC : 0.9054647689, PRC : 0.8843049327, RCL : 0.7068100358, F1 : 0.7856573705
credsweeper Other -> TP : 113, FP : 3, TN : 742, FN : 261, FPR : 0.0040268456, FNR : 0.6978609626, ACC : 0.7640750670, PRC : 0.9741379310, RCL : 0.3021390374, F1 : 0.4612244898
[...]

PR #179, opened 2022-08-08
Fix JWT pattern

### LINKED ISSUE #178: False positive for short regular expressions
Many short regular expressions may randomly apear in the hashes, or base64 lines

Example:

https://github.com/tosca/web/blob/master/packages/EntityFramework.5.0.0/tools/EntityFramework.PS3.psd1#L172

```
$ python -m credsweeper --path EntityFramework.PS3.psd1 
rule: JSON Web Token / severity: medium / line_data_list: [line: '# UK3O3RhOJA/u0afRTK10MCAR6wfVVJUVSZQbQpKumFwwJtoAa+h7veyJBw/3DgSY' / line_num: 172 / path: EntityFramework.PS3.psd1 / value: 'eyJBw' / entropy_validation: False] / api_validation: NOT_AVAILABLE / ml_validation: VALIDATED_KEY
```

99.9% sure it's not really a JWT

While it is a JWT example, this can probably also happen with other short regular expressions (e.g. Google API based on `ya29.` prefix)

As for now i see few options to solve it:
1. Add prefix check and forbid matching regexes if they have a lot of alphanumeric stuff directly to the left. Example: `=eyJBw...` - ok, `dasjkfseyJBw` - bad
2. Add minimal value length for some regexes. For example require minimal JWT length 

Maybe there are some different options. Please propose if you'll have any 

I haven't tested it with the CredData metrics yet, just ideas

babenek:
- added left delimiter for the token
- in tests set ml_threshold to 0.5 due test sample was detected as 0.546
fixes #178


credsweeper/rules/config.yaml
```diff
@@ -162,7 +162,7 @@
   severity: medium
   type: pattern
   values:
-  - (?P<value>eyJ[A-Za-z0-9-_=]+(\.[A-Za-z0-9-_.+\/=]+)?)
+  - (?<![0-9a-zA-Z_-])(?P<value>eyJ[A-Za-z0-9-_=]+\.eyJ[A-Za-z0-9-_=]+(\.[A-Za-z0-9-_=]+)?)
```
meanrin, line 165: JWT with a single `eyJ` entry might still be a valid credential

https://github.com/Samsung/CredSweeper/pull/104

meanrin: `?<![0-9a-zA-Z_-])` probably should not be in JWT regular expression itself
We can concatenate it to ALL pattern based expressions, so it would fix all other short expressions

meanrin (changes requested):
This change referring to the #178 issue, but note that it is not a JWT issue on itself
Same issue can happen with all short patterns
So it would be more correct to apply fix to all pattern based regular expressions

credsweeper/rules/config.yaml
babenek: '''(?<![0-9a-zA-Z_-])''' removes case when eyJ is only a substring in long base64 case. even  ```(?<![0-9a-zA-Z])``` might reduce suspicious strings.

tests/samples/false_positive
```diff
@@ -0,0 +1,6 @@
+
+# https://github.com/tosca/web/blob/f2d42d2b3311994edd68133547fcde8973392464/packages/EntityFramework.5.0.0/tools/EntityFramework.PS3.psd1#L172
+# http://calebb.net/ says: JWT is required to have three segments
```
babenek, line 3:
```suggestion
# http://calebb.net/ says: JWT is required to have three segments
# https://jwt.io/
```

credsweeper/rules/config.yaml
babenek, line 165:
```suggestion
  - (?<![0-9a-zA-Z_-])(?P<value>eyJ[A-Za-z0-9-_=]+(\.eyJ[A-Za-z0-9-_=]+)?(\.[A-Za-z0-9-_=]+)?)
```

Issue #116, opened 2022-05-17
Request for enhancement


P1119r1m:
I think that the algorithm of its work should be revised.
For example, the utility reported 157 results.
There are mostly false alarms:
```
TEEC_ERROR_COMMUNICATIONPASSTEEC_ERROR_SHORT_BUFFERANY_OWNER_SETFAILInvoke_1param_valueInOuthexdump[%zu]:TEEC_InitializeContext
Test passedERROR : TEEC_InvokeCommand failed: %s.
Test passed.%s: OpenSession failed: %x.
# Pass Cond   : TEE_SUCCESS
# Pass Cond    : Passed if unit_tzld passes
Test passedERROR : TEEC_InvokeCommand failed: %s.
 # Pass Cond    : Passed if average time.
[swd_tzisr_test] TESTING : Thread 1:0x%08x:PASS:0x%08x
[swd_tzisr_test] TESTING : 0x%08x:PASS:0x%08x
cpuid[swd_tzisr_test] TESTING : Main Thread :0x%08x:PASS:0x%08x
TEEC_ERROR_COMMUNICATIONPASSTEEC_ERROR_SHORT_BUFFERANY_OWNER_SETFAILhexdump[%zu]:TEEC_InitializeContext
Test failedTest passedERROR: attack test failed: '%s' from '%s'.
```

Such "creds" were found in a binary file.
I haven't checked source codes of this repo, but there is a feeling that I will find there something like:
```
strings BINARY_FILE | grep -i "pass"
```

There have also been reports  of this utility reporting "creds" found in the "DRM"ed (read encrypted) documents.

Dear maintainers, please consider the cases of false positives in order to improve its performance.

Thank you!

babenek:
Did you enabled '--ml_validation' ?
Could you provide command line (or how exactly) you had run the tool?
And if possibly - the data which you tested with false-positive credentials (extension of the file?).

meanrin:
> Our organization doesn't use '--ml_validation' option.

Please conciser using `--ml_validation` if possible, as it the main feature differentiating this scanner from other alternatives 

_Validation by [ML model classifier](https://credsweeper.readthedocs.io/en/latest/overall_architecture.html#ml-validation) is used to reduce False Positives (by far), but might increase False negatives and execution time. So –ml_validation is recommended, unless you want to minimize FN._ [doc link](https://credsweeper.readthedocs.io/en/latest/guide.html)

However, we would take a look at working with binaries, thanks for the description!

[...]

PR #183, opened 2022-08-09
gzip format support

babenek:
- gzip format is extracted like zip


credsweeper/app.py
```diff
@@ -299,6 +300,16 @@ def data_scan(self, data_provider: DataContentProvider, depth: int, recursive_li
             except Exception as zip_exc:
                 # too many exception types might be produced with broken zip
                 logger.error(f"{data_provider.file_path}:{zip_exc}")
+
+        elif Util.is_gzip(data_provider.data):
+            try:
+                with gzip.open(io.BytesIO(data_provider.data)) as f:
```
kmnls, line 306: Why zipped files are checked one-by-one but gzip is parsed as a whole?

credsweeper/app.py
```diff
@@ -299,6 +300,16 @@ def data_scan(self, data_provider: DataContentProvider, depth: int, recursive_li
...
+                    gzip_content_provider = DataContentProvider(data=f.read(), file_path=data_provider.file_path)
```
kmnls, line 307: No check for excluded files?

credsweeper/app.py
babenek: gzip contains single file only. tar is applied when necessary pack many files in gzip

credsweeper/app.py
babenek: it will be throw through exception. 
excluding extension is removed from list when --depth enabled

credsweeper/config/config.py
```diff
@@ -33,5 +33,8 @@ def __init__(self, config: Dict) -> None:
         self.find_by_ext: bool = config["find_by_ext"]
         self.size_limit: Optional[int] = parse_size(config["size_limit"]) if config["size_limit"] is not None else None
         self.depth: int = int(config["depth"])
-        if self.depth > 0 and ".zip" in self.exclude_extensions:
-            self.exclude_extensions.remove(".zip")
+        if 0 < self.depth:
+            if ".zip" in self.exclude_extensions:
+                self.exclude_extensions.remove(".zip")
+            if ".gz" in self.exclude_extensions:
+                self.exclude_extensions.remove(".gz")
```
babenek, line 40: @kmnls here

kmnls: The solution is inconsistent. If the user provides zip as a file type to skip you should not silently remove this exclude list.
I propose to check zip and gz only if they are not excluded manually.

kmnls: @csh519 You opinion?

babenek:
WOW: https://github.com/Samsung/CredSweeper/runs/7767234941?check_suite_focus=true - coverage decreased. That means many code was added.

credsweeper/config/config.py
csh519: @kmnls 

As you told this solution is inconsistent.. It would be good to there is another way to extract files for zip files(data scan)..

I think `--depth` scanning option is option for CredSweeper, and it means the user agreed to scan zip files.
So it doesn't matter if `.zip`, `.gzip` extensions are in config file.
I propose to add new file extractor for zip file or other possible file types. (Maybe `FileForceExtractor`..?)

credsweeper/app.py
csh519: Regarding excluded files there is `check_exclude_file()` method in zip file case.

https://github.com/Samsung/CredSweeper/blob/70b15c72e3cb8c613a1d2065505706cf7148c4bb/credsweeper/app.py#L288-L289

But not in gzip case..

https://github.com/Samsung/CredSweeper/blob/70b15c72e3cb8c613a1d2065505706cf7148c4bb/credsweeper/app.py#L304-L311

I think it can be solve with `FileForceExtractor`(temporary name..), that I wrote below.

credsweeper/config/config.py
babenek: The feature --depth was introduced after issue #116. It is only partially solution and use only standard libraries. 
Assume, --depth will help research APK files that are not excluded. Thus, ZIP extension is do. 
So, I did the workaround for zip files to prove the solution. BTW, xlsx, docx and so on are zip archives too.
Therefore we have to exclude pure binary files e.g. bmp, jpg and so on. Quickest way is skip by extension.

Proposal 1 (easiest):
- add exclude list for binary files which are (might be) containers (zip, gzip, tar, iso ...) 
- when --depth is not defined - exclude found file from both exclusion lists. 
- when --depth is defined - exclude only from standard list
Proposal 2 :
- add totally file recognition by signature and skip unsupported binary files (e.g. git supposes that file is binary when 0x00 is in first 8000bytes, with some workaround for utf-16...) + something like https://github.com/h2non/filetype.py.
- skip files by extension list as currently if --depth is not enabled;
- read all files and skip binary files which are not supported when --depth enabled

Let the PR will be merged as is and new feature will be created after solution designed.

babenek: Before new issue creation - let decide what way preferred.

kmnls: I propose the following solution.
In the config file we preserve the excluded_extensions section.
To the config file we add our  "depth" option as 2 options "check_inside_archives"+"depth", plus we add additional list "include" (if "zip" format is in then it will be checked).
Thus, if the file is in excludes list then it will be skipped anyway, and if not then:
- if check_inside_archives + in includes then it will be unzipped and checked
- if not chec_inside_archives then it will be scanned as binary.

credsweeper/validations/github_token_validation.py
```diff
@@ -38,7 +38,7 @@ def verify(cls, line_data_list: List[LineData]) -> KeyValidationOption:
                 headers={"Authorization": f"token {line_data_list[0].value}"},
             )
         except (requests.exceptions.ConnectionError, Exception) as exc:
-            logger.info(exc)
+            logger.error(exc)
```
meanrin, line 41: Please describe why do we weed that change, and how does it related to gzip archives mentioned in the PR description

fuzz/__main__.py
```diff
@@ -13,83 +13,129 @@
...
+# need to set depth=1 to remove .zip and .gz from extension
+cred_sweeper = CredSweeper(depth=1)
+api_validation = ApplyValidation()
+
+INPUT_DATA_SIZE = 0x0600
+BEHAVIOUR_BYTE_SIZE = 0x01
+MOCK_RESPONSE_SIZE = 0x01FF
+
+
+def mock_request(behaviour_code: int, status_code_seed: int, content: bytes, candidate, patch_object, path_name):
```
meanrin, line 49: Can you please describe changes been done here, and how they are relates to gzip files?

meanrin (changes requested):
Please provide detailed description on what changes included in that PR and for what reason

And do not bloat your PR, keep our dev policies in mind:
> The following list shows some guiding rules that you should keep in mind when you're contributing to the CredSweeper:
>
> - DON'T surprise us with large pull requests. Instead, file an issue and start a discussion so we can agree on a direction before you invest a large amount of time.

https://github.com/Samsung/CredSweeper/blob/main/docs/howto/how-to-contribute.md#dos-and-donts

I see that some of it relates to gzips, some to API request mocking, and some to logging
While fist change is described, there are no obvious motivation for the last two, or why they are in the same PR at all

credsweeper/app.py
babenek: There is other issue - if --depth is enabled, does the user want to scan obviously archives (zip,gz,...) or not? Probably only APK, or XLSX ...
As you can see --depth reads file anyway to recognize the type. Extension may be various or absent..

babenek: data_provider.file_path - yes, there is no workaround to remove *.gz if exists.

fuzz/__main__.py
babenek: New code has reduced percent of dynamic testing (fuzzing) - i had to refuzzed but found that something might be optimized... #189

credsweeper/validations/github_token_validation.py
babenek: It was found during debug fuzzing  #188

babenek:
Created #192 to discuss concern `--depth` behavior. 
Propose to merge the PR as is and made next PR with disabling file extension check when `--depth` is enabled

meanrin (approved):
LGTM, thank you for changing the PR based on the mentioned concerns

csh519 (approved):
Okay. Let's merge it first.

LGTM :+1:

PR #184, opened 2022-08-09
Left barrier for short patterns which may be found in long base64 data

babenek:
Updated regex to skip small patterns which might be appear in long base64 encoded data.
The approach removes one wrong JWT credentials, looks like encrypted data:
```
 *   ciphertext: '6GeyJOCMluH-........
```
Updated experiment with the benchmark scores


csh519 (commented):
If the left barrier needed.. how about the right barrier?
Could you please test about this?

babenek:
@csh519 right barrier cannot be applied for various length patterns. '+' - is greedy and gets all symbols. 
I'll try check right barrier for fixed size only .

babenek:
@csh519 https://github.com/Samsung/CredSweeper/pull/184/commits/e8eeae92c81ee1cb59c7cf5207a4e72c784b7e37
- only tests are updated due Square-secrets was recognized as Square-ID
- benchmark still the same.

csh519 (approved):
Thank you for improve patterns.

LGTM 👍

kmnls (approved):

PR #185, opened 2022-08-10
Update action.yml

babenek:
- check minimal count of credentials to verify common functionality to avoid often issue when main branch has  different samples count, rules and so on.


.github/workflows/action.yml
```diff
@@ -34,7 +34,7 @@ jobs:
     - name: Prepare commit status
       run: |
         ITEMS_CNT=$(jq '. | length' output.json)
-        if [ 18 -eq ${ITEMS_CNT} ]; then
+        if [ 12 -lt ${ITEMS_CNT} ]; then
```
babenek, line 37:
```suggestion
        if [ 1 -lt ${ITEMS_CNT} ]; then
```

iuriimet (approved):

csh519 (approved):
Okay, just check the CredSweeper executed successfully with minimal 1 result.

LGTM 👍

PR #187, opened 2022-08-10
Consolidated json load-dump methods

babenek:
- json is loaded and dumped  through  Util method in all project. Purpose: reuse in import-export config
- added tests for encoding utf-16 and utf-8


meanrin (changes requested):
Strongly oppose to this idea
`.onnx` and `.json` model files is not the python code, and should not be treated as such 
This change is akin moving `.jpg` or `.dll` files to inside the python code 
It’s just much more complicated and obscure 

Moreover, that change would make it uncomfortable to change model files in case of retrain
Currently it is relatively easy to substitute a new model if we trained one: copy-paste 2 files from train directory. But with this change it would require to rewrite python code and use additional base 64 conversion

babenek:
@meanrin Ok, let the ml model data left as is.
I'm having intend to move some config to python code to export for user modification and import. #170 
I hope, you have no objection?

credsweeper/utils/util.py
```diff
@@ -267,3 +269,22 @@ def read_data(path: str) -> Optional[bytes]:
         except Exception as exc:
             logger.error(f"Unexpected Error: Can not read '{path}'. Error message: '{exc}'")
         return None
+
+    @staticmethod
+    def json_load(file_path: str, encoding=DEFAULT_ENCODING) -> Any:
```
meanrin, line 274: Ooo, fancy
Thank you for adding those functions

tests/utils/test_util.py
```diff
@@ -274,3 +274,93 @@ def test_is_gzip_n(self):
...
+    def test_json_load_n(self):
+        self.assertIsNone(Util.json_load("not_existed_path"))
+        with tempfile.TemporaryDirectory() as tmp_dir:
+            self.assertTrue(os.path.isdir(tmp_dir))
+            file_path = os.path.join(tmp_dir, __name__)
+            with open(file_path, "wb") as f:
+                f.write(AZ_DATA)
+            self.assertIsNone(Util.json_load(file_path))
+
+    def test_json_dump_p(self):
```
meanrin, line 332: Imo it's a bit unclear what is being tested here
Docstring comments might be a nice idea
please conciser adding them with that or different PR

meanrin (approved):
LGTM
However, I think we need to add more comments to tests as they might be less obvious with each new functionality added

kmnls (approved):

PR #188, opened 2022-08-11
Raise exception log level to ERROR in api validation

babenek:
During api validation there might be exceptions with communication. Usually it is not displayed due log level is warning by default.


csh519 (commented):
What about `warning`?
I think it is not that important in CredSweeper behavior.

meanrin:
> What about `warning`? I think it is not that important in CredSweeper behavior.

@csh519 while not important it's still error like behavior. If person decided to run `--api_validation` they likely want to know if it's failed

I think most likely cause for such behavior is:
1. No internet connection 
2. ddos related ban or DNS/Proxy issues 

@babenek how about we change 
`logger.error(exc)`
to smth like
`logger.error("Cannot validate XYZ token using API. No connection to the server or wrong response.")`



IMO good change in a concept, but i think the message itself can be more concrete (example above)

csh519:
> > What about `warning`? I think it is not that important in CredSweeper behavior.
> 
> @csh519 while not important it's still error like behavior. If person decided to run `--api_validation` they likely want to know if it's failed
> 
> I think most likely cause for such behavior is:
> 
> 1. No internet connection
> 2. ddos related ban or DNS/Proxy issues
> 

That sounds fine to me.

babenek:
There might be many variations of exception type that we cannot solve. 
So, i'he added common error description.

kmnls (approved):

csh519 (approved):
API validation exceptions are changed to raise error with common description.

LGTM 👍

PR #189, opened 2022-08-11
Fuzz target refactored, corpus refuzzed

babenek:
- optimized fuzz consumer;
- reduced mocking - only for specific Candidate
- instumentation is enabled only when environment variable DO_ATHERIS_INSTRUMENT is set - more convenience for debug.
- increased up to 78% and added seeds for gzip fuzzing #183 
- refuzzed corpus


fuzz/__main__.py
```diff
@@ -13,83 +13,129 @@
...
+            api_validation.validate(candidate)
+    if 1 == behaviour_code:
+        logger.debug(f"<<<<<<<< InvalidGrantError('fuzz InvalidGrantError')")
+        with patch.object(InstalledAppFlow, InstalledAppFlow.from_client_config.__name__) as mock:
+            flow = MagicMock()
+            flow.fetch_token.side_effect = InvalidGrantError('fuzz InvalidGrantError')
+            mock.return_value = flow
+            api_validation.validate(candidate)
+    else:
+        logger.debug(f"<<<<<<<< Exception('fuzz flow Exception')")
```
meanrin, line 88: Maybe replace `<<<<<<<<` by some kind of tag? 
It looks kinda messy

fuzz/__main__.py
```diff
@@ -13,83 +13,129 @@
...
+                    mock_flow(behaviour_code, candidate)
 
     cred_sweeper.credential_manager.set_credentials(candidates)
     cred_sweeper.post_processing()
 
 
 def main():
-    # # # Instrument all works with ~20K functions, but it does not lose seeds during reducing
-    if not os.getenv('SKIP_ATHERIS_INSTRUMENT'):
+    # # # Instrument all works with ~26K functions, but it does not lose seeds during reducing
```
meanrin, line 137: Three `#` looks very fancy

meanrin (approved):
Likely LGTM
However, I'll be honest: have no idea what exactly fuzz/corpus is doing, even after reading that comment https://github.com/Samsung/CredSweeper/pull/107#issuecomment-1115995194 

@babenek do we have fuzzing related documentation in the readme or docs? If not, it would be nice to have few sentences describing it's behavior

fuzz/__main__.py
babenek: The trick helps to avoid issue with automatic uncomment lines in IDE when real comment line is selected by mistake.

fuzz/__main__.py
babenek: Probably.
Otherwise, simple graphics is easiest to find than text recognition by human (IMHO). There were many text in debug, but this samples did not appear in data. So, it was choose like a marker to distinguish in(>>>>>) and out(<<<<<).

babenek:
@meanrin Dynamic code analysis is applied to met criteria https://bestpractices.coreinfrastructure.org/en/projects/6055#analysis
fuzz/corpus is used to store data which was collected after fuzzing. It is like unit-testing, but with possibly find something unexpected like in API validation.
https://github.com/Samsung/CredSweeper/blob/main/fuzz/README.md

kmnls (approved):

PR #191, opened 2022-08-17
Custom config for CredSweeper

babenek:
- export default config to file. User might change the config.
- use custom config from user path
- tests
MERGE AFTER #187 
concerning #69

### DIFF
--- a/tests/test_app.py
+++ b/tests/test_app.py
@@ -343,7 +343,8 @@ class TestApp(TestCase):
     def test_import_config_n(self) -> None:
         # not existed file
         _stdout, _stderr = self._m_credsweeper(
-            ["--config", "not_existed_file", "--path", str(CREDSWEEPER_DIR), "--log", "CRITICAL"])
+            ["--config", "not_existed_file", "--path",
+             str(CREDSWEEPER_DIR), "--log", "CRITICAL"])
         self.assertEqual(0, len(_stderr))
         self.assertIn("CRITICAL", _stdout.decode())
         # wrong config
@@ -352,7 +353,8 @@ class TestApp(TestCase):
             with open(json_filename, "w") as f:
                 f.write('{}')
             _stdout, _stderr = self._m_credsweeper(
-                ["--config", json_filename, "--path", str(CREDSWEEPER_DIR), "--log", "CRITICAL"])
+                ["--config", json_filename, "--path",
+                 str(CREDSWEEPER_DIR), "--log", "CRITICAL"])
             self.assertEqual(0, len(_stderr))
             self.assertIn("CRITICAL", _stdout.decode())
 
--- a/tests/test_app.py
+++ b/tests/test_app.py
@@ -316,7 +316,7 @@ class TestApp(TestCase):
         with tempfile.TemporaryDirectory() as tmp_dir:
             custom_config = os.path.join(tmp_dir, f"{__name__}.json")
             shutil.copyfile(CREDSWEEPER_DIR / "secret" / "config.json", custom_config)
-            args = ["--config", custom_config, "--path", CREDSWEEPER_DIR, "--find-by-ext", "--log", "CRITICAL"]
+            args = ["--config", custom_config, "--path", str(CREDSWEEPER_DIR), "--find-by-ext", "--log", "CRITICAL"]
             _stdout, _stderr = self._m_credsweeper(args)
             self.assertEqual("", _stderr.decode())
             output = _stdout.decode()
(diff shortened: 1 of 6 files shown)


kmnls (dismissed):

tests/test_app.py
```diff
@@ -280,25 +283,52 @@ def test_version_p(self) -> None:
...
 
     def test_patch_save_json_n(self) -> None:
         target_path = str(SAMPLES_DIR / "password.patch")
         _stdout, _stderr = self._m_credsweeper(["--diff_path", target_path, "--log", "silence"])
-        assert not os.path.exists(os.path.join(PROJECT_DIR, "unittest_output_added.json"))
-        assert not os.path.exists(os.path.join(PROJECT_DIR, "unittest_output_deleted.json"))
+        for root, dirs, files in os.walk(PROJECT_DIR):
+            self.assertIn('credsweeper', dirs)
+            for file in files:
+                self.assertFalse(file.endswith('.json'))
```
csh519, line 300: It is not a issue.. but if you don't mind could you please change the `'` to `"`?

tests/test_app.py
```diff
@@ -280,25 +283,52 @@ def test_version_p(self) -> None:
...
+            break
```
csh519, line 301: In my opinion using break and this for loop is not that improve the code than before.
How do you think about it?

tests/test_app.py
```diff
@@ -280,25 +283,52 @@ def test_version_p(self) -> None:
...
+            assert os.path.exists(json_filename)
+
+    # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # #
+
+    def test_import_config_p(self) -> None:
+        with tempfile.TemporaryDirectory() as tmp_dir:
+            json_filename = os.path.join(tmp_dir, f"{__name__}.json")
+            shutil.copyfile(CREDSWEEPER_DIR / "secret" / "config.json", json_filename)
+            _stdout, _stderr = self._m_credsweeper(["--config", json_filename, "--log", "silence"])
+            assert os.path.exists(json_filename)
```
csh519, line 318: I think asserting `json_filename` is good, but it would be better to check the config file imported properly.
How about check `self.config` variable from `app.py` is initialized successfully?

babenek: The test verifies whether current config might be loaded like custom config.
No direct way to check self.config due credsweeper is run in separated process.
I'll try to make more sense in the test

csh519: @babenek 

I know it is difficult too.. Thank you for your trying :+1:

babenek: Done. There are .py files are found with the extension adding in custom config

tests/test_app.py
```diff
@@ -280,25 +285,84 @@ def test_version_p(self) -> None:
...
+        with tempfile.TemporaryDirectory() as tmp_dir:
+            custom_config = os.path.join(tmp_dir, f"{__name__}.json")
+            shutil.copyfile(CREDSWEEPER_DIR / "secret" / "config.json", custom_config)
+            args = ["--config", custom_config, "--path", str(CREDSWEEPER_DIR), "--find-by-ext", "--log", "CRITICAL"]
+            _stdout, _stderr = self._m_credsweeper(args)
+            self.assertEqual("", _stderr.decode())
+            output = _stdout.decode()
+            self.assertNotIn("CRITICAL", output)
+            self.assertIn("Detected Credentials: 0", output)
+            self.assertIn("Detected Credentials: 0", output)
```
csh519, line 325: Duplicate `assertIn()` checks.
Please check.

tests/test_app.py
```diff
@@ -280,25 +285,84 @@ def test_version_p(self) -> None:
...
+    def test_import_config_p(self) -> None:
+        with tempfile.TemporaryDirectory() as tmp_dir:
+            custom_config = os.path.join(tmp_dir, f"{__name__}.json")
+            shutil.copyfile(CREDSWEEPER_DIR / "secret" / "config.json", custom_config)
+            args = ["--config", custom_config, "--path", str(CREDSWEEPER_DIR), "--find-by-ext", "--log", "CRITICAL"]
+            _stdout, _stderr = self._m_credsweeper(args)
+            self.assertEqual("", _stderr.decode())
+            output = _stdout.decode()
+            self.assertNotIn("CRITICAL", output)
+            self.assertIn("Detected Credentials: 0", output)
```
babenek, line 324:
```suggestion
            self.assertIn("Time Elapsed:", output)
```

tests/test_app.py
babenek: Thank you.  "Time Elapsed:" is checked too

tests/test_app.py
```diff
@@ -280,25 +285,84 @@ def test_version_p(self) -> None:
...
+            modified_config = Util.json_load(custom_config)
+            self.assertIn("find_by_ext_list", modified_config.keys())
+            self.assertIsInstance(modified_config["find_by_ext_list"], list)
+            modified_config["find_by_ext_list"].append(".py")
+            Util.json_dump(modified_config, custom_config)
+            _stdout, _stderr = self._m_credsweeper(args)
+            output = _stdout.decode()
+            self.assertEqual("", _stderr.decode())
+            self.assertNotIn("CRITICAL", output)
+            self.assertIn("Detected Credentials:", output)
```
csh519, line 337: If then I think below code block should be changed like below.

```suggestion
            self.assertIn("Time Elapsed:", output)
```

csh519 (approved):
Custom config import and export functions are implemented.

LGTM :+1:

kmnls (approved):

PR #196, opened 2022-08-19
Removed extra file

babenek:
removed forgotten binary file


csh519 (approved):
How does it included? Don't we using `.gitignore`?

meanrin (approved):
Thank you for deleting it

babenek:
> How does it included? Don't we using `.gitignore`?

all corpus directory is ignored and there was a  mistake when used git add --force fuzz/

PR #197, opened 2022-08-21
Moved keyword_checklist into config

### LINKED ISSUE #195: Filter out false positives
There is an option in config.json to exclude pattern and the README file includes a credential example. But this option in code is used to filter by file name, it does not filter credentials.

There should be a way to filter out patterns that are known to be public keys or non-sensitive data of the project.

babenek:
- moved keyword checklist from separated text file to config


credsweeper/rules/config.yaml
```diff
@@ -13,7 +13,7 @@
   severity: high
   type: pattern
   values:
-  - (^|[^0-9A-Za-z])(?P<value>(ABIA|ACCA|AGPA|AIDA|AIPA|AKIA|ANPA|ANVA|AROA|APKA|ASCA|ASIA)[0-9A-Z]{16,17})([^0-9A-Za-z]|$)
+  - (^|[^0-9A-Za-z])(?P<value>(ABIA|ACCA|AGPA|AIDA|AIPA|AKIA|ANPA|ANVA|AROA|APKA|ASCA|ASIA)(?!ANDPACIFICREGION)[0-9A-Z]{16,17})([^0-9A-Za-z]|$)
```
babenek, line 16:
```suggestion
  - (^|[^0-9A-Za-z])(?P<value>(ABIA|ACCA|AGPA|AIDA|AIPA|AKIA|ANPA|ANVA|AROA|APKA|ASCA|ASIA)[0-9A-Z]{16,17})([^0-9A-Za-z]|$)
```

credsweeper/filters/value_dictionary_keyword_check.py
```diff
@@ -21,6 +22,6 @@ def run(self, line_data: LineData) -> bool:
         """
         if line_data.value is None:
             return True
-        if any(keyword in line_data.value.lower() for keyword in self.keyword_checklist.get_list()):
+        if any(keyword == line_data.value.lower() for keyword in self.keyword_checklist):
```
babenek, line 25: Suppose, there should be equals, but many tests fail wit growing number of credentials.
@csh519, @meanrin what do you think?

credsweeper/secret/config.json
```diff
@@ -1,150 +1,1097 @@
 {
-    "exclude": {
-        "pattern": [],
-        "extension": [
-            ".7z",
-            ".aac",
-            ".aar",
-            ".avi",
```
csh519, line 8: In my VS Code 4 spaces is default value of formatting..
Are you using Pycharm?

credsweeper/secret/config.json
```diff
@@ -1,150 +1,1097 @@
...
+  ],
+  "keyword_checklist": [
+    "1234",
+    "abort",
+    "about",
+    "above",
+    "absolute",
+    "abstract",
+    "accent",
+    "accept",
```
csh519, line 158: Adding `keyword_checklist` makes `config.json` file too long.. isn't it?
This is main reason that I created the `txt` file for it.

credsweeper/secret/config.json
babenek: Yes. it was default in Pycharm. I replaced to 4 spaces.

credsweeper/secret/config.json
meanrin: And I want to remind that it's more difficult to add values to json than plaintext

I think all blacklists I remember always was stored as a plaintext, here few examples: 
https://github.com/danielmiessler/SecLists/tree/master/Passwords
https://github.com/berandal666/Passwords
https://github.com/Python-Fa/Password-List

babenek: Ok. May i remove only keyword_checklist.py and use simply logit to load the txt file?
And could you review draft https://github.com/Samsung/CredSweeper/pull/205 ? - the issue i found when working with the checklist.,

PR #198, opened 2022-08-22
Update ml related documentation

meanrin:
- update model architecture
- add example with detailed steps used by the ML model
- mention that ML enabled by the default
- add info about thresholds

Notes:
- You will see change in image only after PR as it use full link to the raw.githubusercontent.com
- Some links use commit hash at url (e.g. `https://github.com/Samsung/CredSweeper/blob/6a2e575987448dd20895a8e72efb3b09fdcbecc2/credsweeper/ml_model/ml_validator.py#L29`). This is done so it will point to the correct object even if we would add/remove few lines in the latest release. User may change to the current main branch and locate corresponding line by themselves


babenek (approved):
OK. Documentation is updated only

csh519 (approved):
Updating documents always good.

LGTM :+1:

PR #202, opened 2022-08-30
Benchmark workflow

babenek:
- Benchmark workflow for calculate scores. Using github cache reduces time to download CredData.
- The workflow fails when scores are different. Need to update cicd/benchmark.txt.
- Used v3 in all github actions
- updated how-to-contribute
- removed ActionTest from PR but keep in push trigger.


meanrin (commented):
Useful feature, LGTM

How about adding description about it to the readme or dev policies?
Some thing like "you need to check results produced by benchmark. If they are different please update XXX file and report changes to PR description"

csh519:
It is cool feature..!

However is it possible to store our benchmark data files to Github server..?
I think it is too huge for Github..

babenek:
> It is cool feature..!
> 
> However is it possible to store our benchmark data files to Github server..? I think it is too huge for Github..

[10Gb](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy) is available but our data is only 1Gb

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,125 @@
...
+
+name: Benchmark
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+  schedule:
+    - cron: '0 2 * * *'
```
babenek, line 12:
```suggestion
```

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,125 @@
...
+        grep '^Detected Credentials: \|^result_cnt : \|^credsweeper ' credsweeper.log | tee benchmark.txt
+
+    - name: Upload artifact
+      uses: actions/upload-artifact@v3
+      with:
+        name: benchmark
+        path: benchmark.txt
+
+    - name: Verify benchmark scores of the PR
+      # update cicd/benchmark.txt with uploaded artifact
```
babenek, line 122:
```suggestion
      # update cicd/benchmark.txt with uploaded artifact if a difference is found
```

csh519:
> [10Gb](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy) is available but our data is only 1Gb

@babenek 

Thank you for explanation.
BTW following the descriptions of [limitation policy](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy) Github remove the cached data which have not been accessed over 7 days.
So in my opinion we should to run this workflow everyday for cached data. How do you think about it?

babenek:
@csh519 
I think, we do PR more often :)
In worst case - the action runs +20 min longer.

csh519:
> @csh519 I think, we do PR more often :) In worst case - the action runs +20 min longer.

@babenek 

Yeah that's my utopia. :smile: 
But as you know in the worst case we need more than 20 min and even the data can be different.. (Removed repository or make private or so on..)
So I worried about the correctness of benchmark result that due to data losing.

babenek:
@csh519 
Yes, data may be lost, e.g. TypeScript-Node-Starter removed interested commit :) so, we [have to removed](https://github.com/Samsung/CredData/pull/19/files) the repo.
Anyway, we can test PR scores diff in the case.
When data is changed - need to test unchanged CredSweeper, update new scores in main and then - run benchmark in new PR.

csh519:
> When data is changed - need to test unchanged CredSweeper, update new scores in main and then - run benchmark in new PR.

@babenek 

From this perspective, it makes sense.

csh519 (approved):
Adding benchmark test in Github workflow.

LGTM :+1:

meanrin (approved):
LGTM, thank you for adding that workflow

PR #204, opened 2022-08-31
Test benchmark

meanrin:
Just a test to check benchmark cache


babenek (commented):
Thanks!
Hypothesis: the cache is kept when ONE or FIRST pull request is opened

PR #207, opened 2022-08-31
Add option to exclude outputs with cli and config

### LINKED ISSUE #195: Filter out false positives

meanrin:
## Description

Resolves #195 

- add blacklist field to the default config. Old configs without the fields will still be processed normally
- add separate value and line blacklists. Both ignores space-like characters at left and right
- add blacklist CLI arg that extends the list from config
- at the moment CLI blacklist adds to both line and value blacklist for simplicity and user convenience. However, interface allows to add line and value lists separately 
- blacklist stored as Sets rather than Lists. Sets on average faster for `x in y` checks, and have average search time of O(1) https://wiki.python.org/moin/TimeComplexity 
- update the guide with blacklisting details 

image

## How has this been tested?

- [x] Add new unit tests for both CLI arg and passing arguments directly into `CredSweeper`
- [x] Verified proper changes in the guide

### DIFF
--- a/credsweeper/__main__.py
+++ b/credsweeper/__main__.py
@@ -90,10 +90,10 @@ def get_arguments() -> Namespace:
                         default=None,
                         dest="config_path",
                         metavar="PATH")
-    parser.add_argument("--blacklist",
+    parser.add_argument("--denylist",
                         help="path to a plain text file with lines or secrets to ignore",
                         default=None,
-                        dest="blacklist_path",
+                        dest="denylist_path",
                         metavar="PATH")
     parser.add_argument("--find-by-ext",
                         help="find files by predefined extension.",
@@ -203,10 +203,10 @@ def scan(args: Namespace, content_provider: FilesProvider, json_filename: Option
 
     """
     try:
-        if args.blacklist_path is not None:
-            blacklist = [line for line in Util.read_file(args.blacklist_path) if line]
+        if args.denylist_path is not None:
+            denylist = [line for line in Util.read_file(args.denylist_path) if line]
         else:
-            blacklist = []
+            denylist = []
 
         credsweeper = CredSweeper(rule_path=args.rule_path,
                                   config_path=args.config_path,
@@ -219,8 +219,8 @@ def scan(args: Namespace, content_provider: FilesProvider, json_filename: Option
                                   find_by_ext=args.find_by_ext,
                                   depth=args.depth,
                                   size_limit=args.size_limit,
-                                  blacklist_lines=blacklist,
-                                  blacklist_values=blacklist)
+                                  exclude_lines=denylist,
+                                  exclude_values=denylist)
         return credsweeper.run(content_provider=content_provider)
     except Exception as exc:
         logger.critical(exc, exc_info=True)
--- a/credsweeper/__main__.py
+++ b/credsweeper/__main__.py
@@ -204,9 +204,7 @@ def scan(args: Namespace, content_provider: FilesProvider, json_filename: Option
     """
     try:
         if args.blacklist_path is not None:
-            with open(args.blacklist_path) as f:
-                blacklist_text = f.read()
-            blacklist = [line for line in blacklist_text.split("\n") if line]
+            blacklist = [line for line in Util.read_file(args.blacklist_path) if line]
         else:
             blacklist = []
 
--- a/credsweeper/__main__.py
+++ b/credsweeper/__main__.py
@@ -90,6 +90,11 @@ def get_arguments() -> Namespace:
                         default=None,
                         dest="config_path",
                         metavar="PATH")
+    parser.add_argument("--blacklist",
+                        help="path to a plain text file with lines or secrets to ignore",
+                        default=None,
+                        dest="blacklist_path",
+                        metavar="PATH")
     parser.add_argument("--find-by-ext",
                         help="find files by predefined extension.",
                         dest="find_by_ext",
@@ -198,6 +203,13 @@ def scan(args: Namespace, content_provider: FilesProvider, json_filename: Option
 
     """
     try:
+        if args.blacklist_path is not None:
+            with open(args.blacklist_path) as f:
+                blacklist_text = f.read()
+            blacklist = [line for line in blacklist_text.split("\n") if line]
+        else:
+            blacklist = []
+
         credsweeper = CredSweeper(rule_path=args.rule_path,
                                   config_path=args.config_path,
                                   api_validation=args.api_validation,
@@ -208,7 +220,9 @@ def scan(args: Namespace, content_provider: FilesProvider, json_filename: Option
                                   ml_threshold=args.ml_threshold,
                                   find_by_ext=args.find_by_ext,
                                   depth=args.depth,
-                                  size_limit=args.size_limit)
+                                  size_limit=args.size_limit,
+                                  blacklist_lines=blacklist,
+                                  blacklist_values=blacklist)
         return credsweeper.run(content_provider=content_provider)
     except Exception as exc:
         logger.critical(exc, exc_info=True)
--- a/credsweeper/app.py
+++ b/credsweeper/app.py
@@ -52,7 +52,9 @@ class CredSweeper:
                  ml_threshold: Union[float, ThresholdPreset] = ThresholdPreset.medium,
                  find_by_ext: bool = False,
                  depth: int = 0,
-                 size_limit: Optional[str] = None) -> None:
+                 size_limit: Optional[str] = None,
+                 blacklist_lines: Optional[List[str]] = None,
+                 blacklist_values: Optional[List[str]] = None) -> None:
         """Initialize Advanced credential scanner.
 
         Args:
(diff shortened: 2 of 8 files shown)


credsweeper/__main__.py
```diff
@@ -198,6 +203,13 @@ def scan(args: Namespace, content_provider: FilesProvider, json_filename: Option
 
     """
     try:
+        if args.blacklist_path is not None:
+            with open(args.blacklist_path) as f:
+                blacklist_text = f.read()
+            blacklist = [line for line in blacklist_text.split("\n") if line]
```
babenek, line 209: Need test UTF-16 and windows line ending.

credsweeper/__main__.py
```diff
@@ -198,6 +203,13 @@ def scan(args: Namespace, content_provider: FilesProvider, json_filename: Option
 
     """
     try:
+        if args.blacklist_path is not None:
+            with open(args.blacklist_path) as f:
```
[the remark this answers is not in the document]
meanrin: Is is not overwriting config but works with it together
So person may use both custom config and custom plain text blacklist 

And reading is conditional on `args.blacklist_path` being specified anyway

credsweeper/__main__.py
meanrin: Modified with 0a7f3d85f4e5df14eac1dab809fed578b6df4d3d

csh519 (commented):
Adding those excluding list import code is good!
However how about change the name of `blacklist` to `denylist`.
As you know these days, some words that using for developing is changing.. for "PC" :roll_eyes:

babenek:
> Adding those excluding list import code is good! However how about change the name of `blacklist` to `denylist`. As you know these days, some words that using for developing is changing.. for "PC" roll_eyes

"ignorelist" ???

meanrin:
> "ignore_list" ???

https://english.stackexchange.com/questions/51088/alternative-terms-to-blacklist-and-whitelist

Deny list is the top comment on stackexchange, so here we go

image

credsweeper/__main__.py
```diff
@@ -208,7 +218,9 @@ def scan(args: Namespace, content_provider: FilesProvider, json_filename: Option
                                   ml_threshold=args.ml_threshold,
                                   find_by_ext=args.find_by_ext,
                                   depth=args.depth,
-                                  size_limit=args.size_limit)
+                                  size_limit=args.size_limit,
+                                  exclude_lines=denylist,
+                                  exclude_values=denylist)
```
babenek, line 223: deny_lines
deny_values
- maybe use the 'deny'' ? It helps to keep semantic in code.

meanrin: But we have exclude in config T-T
So it looks very nice when assigning to the config 

image

credsweeper/app.py
```diff
@@ -88,6 +92,10 @@ def __init__(self,
         config_dict["find_by_ext"] = find_by_ext
         config_dict["size_limit"] = size_limit
         config_dict["depth"] = depth
+        if exclude_lines is not None:
+            config_dict["exclude"]["lines"] = config_dict["exclude"].get("lines", []) + exclude_lines
+        if exclude_values is not None:
+            config_dict["exclude"]["values"] = config_dict["exclude"].get("values", []) + exclude_values
```
babenek, line 98: Probably the feature is overhead

credsweeper/config/config.py
```diff
@@ -41,3 +43,7 @@ def __init__(self, config: Dict) -> None:
                 self.exclude_extensions.remove(".zip")
             if ".gz" in self.exclude_extensions:
                 self.exclude_extensions.remove(".gz")
+
+        # Trim exclude patterns from space like characters
+        self.exclude_lines = set(line.strip() for line in self.exclude_lines)
+        self.exclude_values = set(line.strip() for line in self.exclude_values)
```
babenek, line 49: denylist is the same

credsweeper/__main__.py
babenek: with single 'denylist' we will have double work. 
So, i propose use only single denylist in config

credsweeper/scanner/scan_type/scan_type.py
```diff
@@ -154,11 +154,16 @@ def _get_candidate(cls, config: Config, rule: Rule, target: AnalysisTarget) -> O
...
         """
+        if target.line.strip() in config.exclude_lines:
+            return None
+
         line_data = cls.get_line_data(config, target.line, target.line_num, target.file_path, rule.patterns[0],
                                       rule.filters)
 
         if line_data is None:
             return None
+        if line_data.value.strip() in config.exclude_values:
```
babenek, line 165: double work in case of user 'denylist'

babenek (commented):
IMO 'denylist' is duplicated to exclude.lines and exlude.values and it breaks semantic in code (even it described in doc). 
In config it use separated values (OK), but when user applies denylist - double work appears.
User may export default config, modify it with new feature and use like custom config - it will works without extra argument.

credsweeper/scanner/scan_type/scan_type.py
meanrin: > double work in case of user 'denylist'

Not the case
If we stripped values in list, we still need to strip actual lines, otherwise match cannot accurate

meanrin:
> User may export default config, modify it with new feature and use like custom config - it will works without extra argument.

@babenek It would be just a bad UX
I would remind you to try place yourself on the users place, and try to update json file with a new blacklist

credsweeper/scanner/scan_type/scan_type.py
```diff
@@ -154,11 +154,16 @@ def _get_candidate(cls, config: Config, rule: Rule, target: AnalysisTarget) -> O
             remove current line. None otherwise
 
         """
+        if target.line.strip() in config.exclude_lines:
```
babenek, line 157: then, here may be extra work when user suggests skip values only

meanrin: Ok, agree, will update

babenek:
> > User may export default config, modify it with new feature and use like custom config - it will works without extra argument.
> 
> @babenek It would be just a bad UX I would remind you to try place yourself on the users place, and try to update json file with a new blacklist

then maybe there two new arguments?
--denylines
--denyvalues

meanrin:
> > > User may export default config, modify it with new feature and use like custom config - it will works without extra argument.
> > 
> > 
> > @babenek It would be just a bad UX I would remind you to try place yourself on the users place, and try to update json file with a new blacklist
> 
> then maybe there two new arguments? --denylines --denyvalues

Yeah, I thought about it, but i see no case from the user side to be honest

From the dev site i see for the debug reasons OR while training ML

On the user side, collisions between full line and value should be implausible. Therefore, there are should be no random FN due to line be removed because of value other way around

credsweeper/scanner/scan_type/scan_type.py
meanrin: @babenek fixed with 3670e584232a25ef4b2a561e381ac184d1b6057a

credsweeper/__main__.py
meanrin: > with single 'denylist' we will have double work. So, i propose use only single denylist in config

@babenek We are taking about O(1) average difficultly at the search operation
https://wiki.python.org/moin/TimeComplexity

babenek: * N lines

credsweeper/app.py
```diff
@@ -52,7 +52,9 @@ def __init__(self,
                  ml_threshold: Union[float, ThresholdPreset] = ThresholdPreset.medium,
                  find_by_ext: bool = False,
                  depth: int = 0,
-                 size_limit: Optional[str] = None) -> None:
+                 size_limit: Optional[str] = None,
+                 exclude_lines: Optional[List[str]] = None,
+                 exclude_values: Optional[List[str]] = None) -> None:
```
babenek, line 57: if the lines inserted before e.g. find_by_ext - then 55line is kept untouched :)

credsweeper/__main__.py
```diff
@@ -198,6 +203,11 @@ def scan(args: Namespace, content_provider: FilesProvider, json_filename: Option
 
     """
     try:
+        if args.denylist_path is not None:
+            denylist = [line for line in Util.read_file(args.denylist_path) if line]
```
babenek, line 207:
```suggestion
            denylist = [line for line in Util.read_file(args.denylist_path) if line.strip()]
```
skip user lines with whitespaces only?

babenek (approved):
Well, let user decide what way should be used to skip credentials. config or denylist

credsweeper/app.py
meanrin: Agree, but it have a chance to brake code that passes arguments by order rather than byname

credsweeper/__main__.py
meanrin: > yes, only 1 empty check

It would be 1 check IF we would perform linear search

However, `x in y` for sets doing hash map based search that is O(1) in average difficulty - meaning it does not depend on the set size 
So adding `""` to set would increase set size by 1, but not the average number of operations during checking `x in y`

In worst case `x in y` would be O(n), so it would increase operations by 1. But it's really an insanely rare worst case

csh519 (approved):
`--denylist` option is added.

LGTM :+1:

PR #212, opened 2022-09-06
Banner with version and CRC32 sum of program files

babenek:
## Description

- Added `--banner` option to distinguish various version in development. CRC32 sum of files might be displayed in banner at program start. ```CredSweeper v1.4.2 crc32:ae1baf99```. It is not an integrity check but fast.
- Use natural order in benchmark first:old, second:new.

## How has this been tested?

- [x] Add `--banner` to arguments. See crc32 value. 
- [x] Check action 'Static analysis and code style' step 'Integrity and version check'

### DIFF
--- a/credsweeper/__init__.py
+++ b/credsweeper/__init__.py
@@ -1,5 +1,3 @@
-import os
-
 from credsweeper.app import CredSweeper
 from credsweeper.common.constants import ThresholdPreset
 from credsweeper.file_handler import ContentProvider, ByteContentProvider, DiffContentProvider, StringContentProvider, \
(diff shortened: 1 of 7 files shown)


credsweeper/__init__.py
```diff
@@ -8,8 +6,6 @@
 from credsweeper.ml_model.ml_validator import MlValidator
 from credsweeper.validations.apply_validation import ApplyValidation
 
-os.environ["TF_CPP_MIN_LOG_LEVEL"] = "2"
-
```
csh519, line 12: As I know this line is needed to avoid warning log..
Could you explain why this line should be removed?

babenek: The environment used to manage log in tensorflow package. Since we use onnxruntime - it is extra.

csh519: Oh.. That's right. Thank you for remove useless lines.

csh519 (approved):
Thank you for adding `--banner` option for checksum.

LGTM :+1:

iuriimet (approved):
LGTM

kmnls (approved):

PR #215, opened 2022-09-14
Argument --depth and workaround

### LINKED ISSUE #192: Argument --depth and workaround
The issue was created due #183 has to  add second workaround to remove `.gz `extension from excluded list. First was `.zip`.
The argument `--depth` enables file analysis not like plain text but with recursion if the file is an archive. 
Thus, class FilePathExtractor works **before** the analysis and removes true zip archives by extension. So, the feature might look like malfunction.
Suppose, -`-depth` will analyse various file types by content and the exclude list will be ignored.
Simultaneously excluding by `not_allowed_path_pattern` and others are working before the list exluding.
https://github.com/Samsung/CredSweeper/blob/eec642f420a3d27ff0e81489e1e01d61abb980bd/credsweeper/file_handler/file_path_extractor.py#L127-L135

babenek:
## Description

fixes #192 

MERGE AFTER #216 !

- Extract extensions of files which might be scanned with option ``--depth`` to "container" list. There is possibility to scan office files too in future.
- Applied ``lower=True`` for Util.get_extension by default. Config must include the extension in lowercase.
- Tests improvements
- Added line number to log
- Modified action steps to prevent forecast exit when next steps might be done. It helps to find all issues instead one-by-one. Multi-jobs may be extra huge work with duplicates (checkout, python config and so on)

## How has this been tested?

- [x] scan zip file with option --depth 33. The archive must contain at least one credential.
- [x] UnitTests have enough cases for this


tests/utils/test_util.py
```diff
@@ -23,6 +23,12 @@ def test_get_extension_n(self):
 
     def test_get_extension_p(self):
         self.assertEqual(".ext", Util.get_extension("tmp.ext"))
+        self.assertEqual(".JPG", Util.get_extension("tmp.JPG"))
+        self.assertEqual(".Aa", Util.get_extension("tmp.Aa"))
+        self.assertEqual(".Ї", Util.get_extension("tmp.Ї"))
+        self.assertEqual(".♡", Util.get_extension("tmp.♡"))
+        self.assertEqual(".ㅋㅅ", Util.get_extension("tmp.ㅋㅅ"))
+        self.assertEqual(".ß", Util.get_extension("tmp.ß"))
```
csh519, line 31: :rofl:

credsweeper/credentials/line_data.py
```diff
@@ -250,7 +250,7 @@ def is_source_file_with_quotes(self) -> bool:
         """
         if not self.path:
             return False
-        if Util.get_extension(self.path) in self.config.source_quote_ext:
+        if Util.get_extension(self.path).lower() in self.config.source_quote_ext:
             return True
         return False
```
csh519, line 255: How about add default option parameter `lower=True` to `Util.get_extension()`?
I think it would be good.

credsweeper/file_handler/file_path_extractor.py
```diff
@@ -130,7 +130,10 @@ def check_exclude_file(cls, config: Config, path: str) -> bool:
             return True
         if any(exclude_path in path for exclude_path in config.exclude_paths):
             return True
-        if Util.get_extension(path) in config.exclude_extensions:
+        file_extension = Util.get_extension(path)
+        if file_extension in config.exclude_extensions:
+            return True
+        if not config.depth and file_extension in config.exclude_containers:
             return True
```
csh519, line 137: And here, is there any reason to just compare `file_extension` without using `lower()`?
In my opinion we can use and manage all extensions with lower case.

babenek: lower() was applied for ``path`` in the function(line#126), so there is some reducing of double_work.

kmnls (approved):

csh519 (approved):
Requested changes are reflected well!

LGTM :+1:

PR #217, opened 2022-09-18
Custom log config

### LINKED ISSUE #69: Add minimal severity, additional rules, log config, and config arguments

babenek:
## Description

fixes #69

- Applied functional of exporting log config and use user's log config. 
- User may specify log directory. By default - in current working dir.

## How has this been tested?

- [x] --export_log_config - writes built-in config to custom path
- [x] --log_config - applies user's log config


credsweeper/secret/log.yaml
```diff
@@ -4,7 +4,7 @@ version: 1
 
 disable_existing_loggers: False
 
-ignore: [tensorflow, git]
```
csh519, line 7: :eyes:

csh519 (approved):
Thank you for adding `--log_config` and `--export_log_config` option.

LGTM :+1:

kmnls (approved):

PR #219, opened 2022-09-20
Refactored unit test to avoid undefined behaviour

babenek:
## Description

- Fix Unit test due mocking fails randomly

## How has this been tested?
- [x] All unit tests must passed

### DIFF
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -110,13 +110,13 @@ class TestMain:
 
     # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # #
 
-    @mock.patch("pandas.DataFrame", return_value=pd.DataFrame(data=[]))
-    def test_save_xlsx_p(self, mock_xlsx_to_excel: Mock()) -> None:
-        cred_sweeper = CredSweeper(xlsx_filename="unittest_output.xlsx")
-        cred_sweeper.run([])
-        mock_xlsx_to_excel.assert_called()
-        assert os.path.exists("unittest_output.xlsx")
-        os.remove("unittest_output.xlsx")
+    def test_save_xlsx_p(self) -> None:
+        with tempfile.TemporaryDirectory() as tmp_dir:
+            test_filename = os.path.join(tmp_dir, "unittest_output.xlsx")
+            assert not os.path.exists(test_filename)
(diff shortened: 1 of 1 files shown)


babenek:
@csh519 
https://github.com/babenek/CredSweeper/actions/runs/3095173522/jobs/5009308772
I think, there is a patch with Mock issue in new libraries. The tests fail only in "release_test" which use latest libraries.
Temp directory is a practice to use determined environment. I mean, first test fails (TestMain.test_save_xlsx_p) and lefts an artifact,  second test (test_report_p ) checks the same artifact (the same name)

meanrin:
> I wondered why this error occurred randomly.. Do you know why? And is it sure to solve this issue with this change? (Create file to temp directory.)

This issue in the importing order 
Mock `@mock.patch("pandas.DataFrame", return_value=pd.DataFrame(data=[]))`  redefines type `pandas.DataFrame` 
Because of that `if not isinstance(df, DataFrame):` line in the `pandas/io/formats/excel.py` fails, as `pandas.DataFrame` not a type - it's a mock
However, if we call `pandas/io/formats/excel.py` before mock, `pandas/io/formats/excel.py` would already import correct `pandas.DataFrame` as import would happen before the mock. As mock would do nothing to already loaded objects 


So bad order:
1. redefine `pandas.DataFrame`
2. `pandas/io/formats/excel.py` import mocked definition 
3. now `if not isinstance(df, DataFrame):` fails

Good order:
1. `pandas/io/formats/excel.py` import mocked definition 
2. mock doesn't do anything, as `pandas.DataFrame` already loaded into memory 
3.  now `if not isinstance(df, DataFrame):` work

meanrin (approved):
@babenek 
Changes provided by this PR is good

However, 
```python
    @mock.patch("pandas.DataFrame", return_value=pd.DataFrame(data=[]))
    def test_save_xlsx_n(self, mock_xlsx_to_excel: Mock()) -> None:
```
this part of code still replaces the original `DataFrame` object definition 
So the error is still there, we just don't see it because no other tests calls to `DataFrame.to_excel` anywhere
If we would add one more test with call to `to_excel` - error will return 


So while this change on itself is good, we also need to mock only `to_excel` function as I proposed in the comment above https://github.com/Samsung/CredSweeper/pull/219#issuecomment-1253416006

PR #224, opened 2022-09-30
Fix test to avoid extra failures during development

babenek:
## Description

- Fix test to avoid extra failures during development from extra .json files as old artifacts

## How has this been tested?

- [x] create any json file in root of repo and run project (main branch fails)


csh519 (approved):
Avoiding test error is good!

And I think it would be good to increase fuzzing coverage in the future to fit with new calculate method.

LGTM :+1:

kmnls (dismissed):

.github/workflows/fuzz.yml
```diff
@@ -78,7 +78,7 @@ jobs:
       run: |
         COVERAGE=$(tail -1 report.txt | awk '{print $6}' | tr --delete '%')
         # additionally check correctness of the value - should be an integer
-        if ! [ 75 -le ${COVERAGE} ]; then
+        if ! [ 74 -le ${COVERAGE} ]; then
```
babenek, line 81:
```suggestion
        if ! [ 75 -le ${COVERAGE} ]; then
```

iuriimet (approved):
Looks good

csh519 (approved):
Approve again.

PR #226, opened 2022-10-03
Test with password case in xml tag

babenek:
## Description

- Add test for xml 

## How has this been tested?
- All  unit test must passed


csh519 (approved):
Adding sample case is good!
But there is problem on fuzzing checks... :thinking:

csh519:
NVM. I checked PR that related with fuzzing on #224.

kmnls (approved):

PR #229, opened 2022-10-07
Fix test help p error

csh519:
## Description

There is an error in `test_help_p()` method which in `tests/test_app.py` when I trying to run pytest with python3.10.x version.
I investigated the error and I found there is a change on `argparse`.
https://docs.python.org/3/whatsnew/3.10.html#argparse

Following the document ***Misleading phrase “optional arguments” was replaced with “options” in argparse help.*** so I replaced "optional arguemnts" to "options" in `test_help_p()`.

- Fix `test_help_p()` error

## How has this been tested?

- [X] Pytest finished without error

### DIFF
--- a/tests/test_app.py
+++ b/tests/test_app.py
@@ -274,7 +274,8 @@ class TestApp(TestCase):
                     continue
                 if started:
                     # There is argparse change on python3.10 to display just "options:"
-                    if line.strip() == "optional arguments:":
+                    if sys.version_info.major == 3 and sys.version_info.minor >= 10 \
+                        and line.strip() == "optional arguments:":
                         text += line.replace("optional arguments:", "options:")
                     else:
                         text += line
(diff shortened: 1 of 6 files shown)


tests/test_app.py
```diff
@@ -273,7 +273,12 @@ def test_help_p(self) -> None:
                     started = True
                     continue
                 if started:
-                    text += line
+                    # There is argparse change on python3.10 to display just "options:"
+                    if sys.version_info.major == 3 and sys.version_info.minor >= 10 \
```
babenek, line 277: Major version is 3 by default :)

babenek (approved):
OK. 
Python 3.10 tests have passed

tests/test_app.py
csh519: Oops..!
I'll change it later.

PR #230, opened 2022-10-13
Add publickey filtering

csh519:
## Description

There were some cases that have `public` or `pubkey` keyword in variable.

``` text
"encryptedPublicSharingKey": "abcde.........."
pdrendpoint_india_base64_pubkey: abcde..........
```

So I suggest to add `public` and `pubkey` to `VariableNotAllowedPatternCheck` filter.
And there are some changes regarding wrong class names and indent.

- Add `public` and `pubkey` keyword to `VariableNotAllowedPatternCheck` filter
- Fix indent
- Fix wrong class names

## How has this been tested?

- [X] Check the FP cases that variables have `public` or `pubkey` are filtered correctly
- [X] Pytest


babenek:
Could you add a test sample which fails a test without the changes but passes with the PR?

babenek (dismissed):
OK. ``publickey="MIGbMBAGByqG....`` now filtered.
Probably .crt extension might be removed from find_by_ext too.

csh519:
> OK. `publickey="MIGbMBAGByqG....` now filtered. Probably .crt extension might be removed from find_by_ext too.

@babenek 

Thank you for suggestion!
I removed the `.crt` extension file from `find_by_ext` by 0a8f48b.

babenek (approved):
OK. Public keys are not suspicious credentials now.

kmnls (approved):

PR #231, opened 2022-10-17
Fix github action to check last release

babenek:
## Description

- applied recursive scan for source folder - bugfix of CI workflow
- increased period for obsolete release to 28 days. The PR is pass, but on next Monday (25th) it should fail. Then we do release and then may reduce the period again.

## How has this been tested?

- [ ] Check manually after 4 weeks

### DIFF
--- a/.github/workflows/rottenness.yml
+++ b/.github/workflows/rottenness.yml
@@ -57,6 +57,6 @@ jobs:
 
     - name: Compare source code of versions
       if: env.TAG_NAME
-      run: diff ${{ env.TAG_NAME }}/credsweeper ${{ github.event.repository.default_branch }}/credsweeper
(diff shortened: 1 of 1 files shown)


babenek:
@csh519 
the check [does](https://github.com/Samsung/CredSweeper/blob/83f814c764e84a6b57a9cdf8c5111ef84980618b/.github/workflows/rottenness.yml#L47) comparison with main branch 
So, PR should pass but on next Monday we will see new failure if no release.

csh519:
> @csh519 the check [does](https://github.com/Samsung/CredSweeper/blob/83f814c764e84a6b57a9cdf8c5111ef84980618b/.github/workflows/rottenness.yml#L47) comparison with main branch So, PR should pass but on next Monday we will see the failure if no new release.

@babenek 
Okay. I think we should release new version on this week.

csh519 (approved):
Increasing the period to 4 weeks is good.

LGTM :+1:

PR #237, opened 2022-10-27
Parse python sources to find full strings even they were split

babenek:
## Description

- Added built-in parser python with ast. Python3.7 workaround applied.
- Refuzzed seeds

## How has this been tested?

- [x] UnitTest
- [x] split credential string in Python code to multiple lines and check with --depth option


iuriimet (approved):
LGTM

credsweeper/file_handler/data_content_provider.py
```diff
@@ -53,6 +54,19 @@ def represent_as_structure(self) -> bool:
             text = self.data.decode(encoding='utf-8', errors='strict')
         except Exception:
             return False
+        # # # Python
+        try:
+            if ";" in text or 2 < text.count("\n"):
```
csh519, line 59: I don't know why you checking `";"` in here.
Python doesn't use semicolon isn't it?

babenek: Semicolon may be used in python to merge source to one line. 
It is a marker that there is something complicated than HelloWorld. Imports must be separated with semicolon or new lines.
```
import urllib.request; print(urllib.request.urlopen("https://github.com/").read())
```

csh519: Thank you for explanation.

PR #241, opened 2022-11-08
Propagate size limit config to PatchProvider

babenek:
## Description

- Use configured param ``size_limit`` in PatchProvider to skip too big diffs which are computed a long time with ``whatthepath``

## How has this been tested?

- [x] UnitTests
- [x] Set size limit less than a diff size and run credsweeper for scan the diff. Warning about oversize should appear.


csh519 (approved):
Over-sized patch diffs will filter by `size_limit`.

LGTM :+1:

kmnls (approved):

PR #243, opened 2022-11-14
Version up to 1.4.5

### LINKED ISSUE #242: Huge diffs might be parsed very slow
Huge diffs with a lot of lines might be scanned with ``--diff`` very slow.
There is [whatthepatch](https://github.com/cscorley/whatthepatch) bottleneck. https://github.com/cscorley/whatthepatch/issues/48
Temporary solution: use fork with the fix https://github.com/babenek/whatthepatch

babenek:
fixes #242 
## Description

- updated requirement whatthepath to 1.0.3
- Add test for huge diff (git patch) parsing
- version up to 1.4.5

## How has this been tested?

- [x] Make a diff with huge number of lines - scan as diff

### DIFF
--- a/.github/pull-request-template.md
+++ b/.github/pull-request-template.md
@@ -1,6 +1,6 @@
 ## Description
 
-Please include a summary of the change and which is is fixed.
+Please include a summary of the change and which is fixed.
 
 - Add A
 - Fix B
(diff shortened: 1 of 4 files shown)


csh519 (approved):
Let's move upt to v1.4.5!

iuriimet (approved):

PR #255, opened 2023-01-29
Support binary patch with whatthepatch v1.0.4

babenek:
## Description

- Add git binary patch support since whatthepatch v1.0.4 provides the data
- Fix parsing the diff type
- Added patch provider into fuzzing and seeds re-fuzzed

## How has this been tested?

- [x] UnitTest


credsweeper/utils/util.py
```diff
@@ -198,6 +198,34 @@ def patch2files_diff(raw_patch: List[str], change_type: str) -> Dict[str, List[D
...
+        rows_data = []
+        if deleted_line_number is None:
+            # indicates line was inserted
+            rows_data.append(DiffRowData(DiffRowType.ADDED, added_line_number, line))
+        elif added_line_number is None:
+            # indicates line was removed
+            rows_data.append(DiffRowData(DiffRowType.DELETED, deleted_line_number, line))
+        else:
+            rows_data.append(DiffRowData(DiffRowType.ADDED_ACCOMPANY, added_line_number, line))
+            rows_data.append(DiffRowData(DiffRowType.DELETED_ACCOMPANY, deleted_line_number, line))
```
csh519, line 226: I've looked at this code and I don't know why it needed.

As I know CredSweeper only use `DiffRowType.ADDED` and `DiffRowType.DELETED` on [`__main__.py` L#285, L#290](https://github.com/Samsung/CredSweeper/blob/main/credsweeper/__main__.py#L285-L290).

Please check this.

babenek: Yes, it is. 
Only "added" and "deleted" are used but in the case we can obtain a case when added and deleted line number are present. The author has added *_ACCOMPANY types. Maybe for debug. It can be  reduced.

I've fuzzed the patch provider (actually, whatthepatch library) and found many issues which lead to uncaught exceptions (if Credsweeper is used as class).

Additionally, DiffRowType was changed to Enum.

csh519: Okay I've checked all changes, and it sounds good to found many issues from `whatthepatch` :)

csh519 (approved):
Binary patch supports added.

LGTM :+1:

iuriimet (approved):

kmnls (approved):

PR #259, opened 2023-02-02
Add SLSA Release step in pypi.yml

zinmoon:
## Description
Add SLSA Release step in pypi.yml

 SLSA release steps is implemented as reusable-workflow in Samsung/SupplyChainAssurance/.github/workflow/python_releasor.yml

 If use it, below files will be uploaded as assets of release
  - whl, tar.gz for pypi
  - an SBOM(spdx type)
  - an Attestation(slsa provenance)

## How has this been tested?
- [x] Check assets of release

### DIFF
--- a/.github/workflows/pypi.yml
+++ b/.github/workflows/pypi.yml
@@ -21,7 +21,7 @@ jobs:
       version: "3.10"
     secrets:
       EXPECTED_REPOSITORY: "${{ secrets.EXPECTED_REPOSITORY }}"
-      ECODETOKEN: "${{ secrets.ECODETOKEN }}"
+      ECODETOKEN: "${{ secrets.ECODE_TOKEN }}"
       GITHUBTOKEN: "${{ secrets.GITHUB_TOKEN }}"
 
   deploy:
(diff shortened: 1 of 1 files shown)


babenek:
Could you explain the abbreviations?

zinmoon:
> 
Please refer to below url for SLSA.
https://slsa.dev/

CreadSweeper can meet Slsa Level3 if you use reusable-workflow (Samsung/SupplyChainAssurance/.github/workflow/python_releasor.yml)
Also get an SBOM file.

However, some configurations of github are required to work with reusable work flow.
I am deciding with the ShinHyung Choi

babenek (dismissed):
Please, fix the failure
```
fatal: could not read Username for 'https:/r/': No such device or address
```

csh519:
@babenek 

It think this [link](https://github.blog/changelog/2021-04-20-github-actions-control-permissions-for-github_token/) is the cause of `Publish to PyPi`'s failure.
As following the [link](https://github.blog/changelog/2021-04-20-github-actions-control-permissions-for-github_token/)..

> The token has `write` permissions to a number of API endpoints except in the case of pull requests from forks which are always `read`. 

So I'm trying to just merge this PR and see what happens.
If there is some trouble we can revert this PR.

csh519 (approved):
See what happens.

PR #263, opened 2023-02-08
Optimization

babenek:
## Description

- Added limits for data which might be represented for deep scan
- Fix UTF-8 surrogates issue during encoding

## How has this been tested?

- [x] UnitTest

### DIFF
--- a/tests/samples/sample.py
+++ b/tests/samples/sample.py
@@ -1,8 +1,12 @@
+#!/bin/env python
 def get_password() -> str:
     password = \
         "WeR15tr0n6"
     return password
 
 
-if "__main__" == __name__:
+TEST = "\uDF42\uD824"
+
+if __name__ == "__main__":
     print(get_password())
+    print(TEST)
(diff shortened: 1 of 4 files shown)


tests/samples/sample.py
```diff
@@ -1,8 +1,12 @@
+#!/bin/env python
 def get_password() -> str:
     password = \
         "WeR15tr0n6"
     return password
 
 
-if "__main__" == __name__:
+TEST = "\uDF42\uD824"
```
csh519, line 8: Why this `TEST` variable added?

babenek: The file is a sample for test. 
When the sample is parsed with python ast with using '--depth' - it fails with encoding.
E.g.: 
```
>>> print ("\uDF42\uD824")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
UnicodeEncodeError: 'utf-8' codec can't encode characters in position 0-1: surrogates not allowed
```
So, i've added the sample to trigger test fail in future. Like a canary test.

csh519: I see. Thank you for explanation.

csh519 (approved):
Code is optimized and UTF-8 issue is fixed.

LGTM 👍

iuriimet (approved):

PR #267, opened 2023-02-13
Credit card number rule and filter

babenek:
## Description

- Added rule to find credit card number in number sequnce
- Added filter to check the number sequence with Luhn algorithm

Probably necessary to apply https://github.com/Samsung/CredData/pull/22 first - then update the scores

## How has this been tested?

- [x] UnitTest
- [x] Benchmark finds the real  credentials

### DIFF
--- a/credsweeper/filters/cred_card_number_check.py
+++ b/credsweeper/filters/cred_card_number_check.py
@@ -1,0 +1,39 @@
+from credsweeper.credentials import LineData
+from credsweeper.filters import Filter
+
+
+class CreditCardNumberCheck(Filter):
+    """Check that value is a credit card number."""
+
+    def run(self, line_data: LineData) -> bool:
+        """Run filter checks on received credential candidate data 'line_data'.
+
+        Args:
+            line_data: credential candidate data
+
+        Return:
+            False, if the sequence is not card number. True if it is
+
+        """
+        if line_data.value is None \
+                or 16 != len(line_data.value) \
+                or ("0" == line_data.value[0] and "0" == line_data.value[1]):
+            return True
+        try:
+            s = 0
+            # https://en.wikipedia.org/wiki/Luhn_algorithm
+            for n in range(0, 16):
+                x = int(line_data.value[n])
+                if 0 == (1 & n):  # Only for odd numbers (with 0 as a start index)
+                    x *= 2
+                    if x > 9:
+                        x -= 9
+                s += x
+
+            if 0 == s % 10:
+                return False
+        except ValueError:
+            pass
+
+        # return False when the sequence is not a credit card number
+        return True
--- a/cicd/benchmark.txt
+++ b/cicd/benchmark.txt
@@ -1,5 +1,5 @@
-Detected Credentials: 4456
-result_cnt : 4062, lost_cnt : 70, true_cnt : 3665, false_cnt : 327
+Detected Credentials: 4464
+result_cnt : 4070, lost_cnt : 78, true_cnt : 3665, false_cnt : 327
 credsweeper -> TP : 3665, FP : 327, TN : 19429556, FN : 910, FPR : 0.0000168297, FNR : 0.1989071038, ACC : 0.9999363502, PRC : 0.9180861723, RCL : 0.8010928962, F1 : 0.8556087312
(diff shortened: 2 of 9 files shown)


credsweeper/filters/cred_card_number_check.py
```diff
@@ -0,0 +1,39 @@
...
+            for n in range(0, 16):
+                x = int(line_data.value[n])
+                if not 1 & n:
+                    x *= 2
+                    if x > 9:
+                        x -= 9
+                s += x
+
+            if 0 == s % 10:
+                return False
```
csh519, line 34: Could you explain it more?
I can't understand even I read the wiki page.
The code block is different with pusedo code from wiki page.

babenek: You might be confused. 
I used the algorithm, where the size is 16 (regex in rule is fixed). e.g. https://www.groundlabs.com/blog/anatomy-of-a-credit-card/
Then, move "from left to right". The wiki pseudocode has opposite direction and the original Luhn algorithm is designed for various size of number.
The approach may be used for 15 symbols cards with zero-left-padding.
E.g. AmericanExpress 378282246310005 - > 0378282246310005

credsweeper/filters/cred_card_number_check.py
```diff
@@ -0,0 +1,39 @@
...
+        if line_data.value is None \
+                or 16 != len(line_data.value) \
+                or ("0" == line_data.value[0] and "0" == line_data.value[1]):
+            return True
+        try:
+            s = 0
+            # https://en.wikipedia.org/wiki/Luhn_algorithm
+            for n in range(0, 16):
+                x = int(line_data.value[n])
+                if not 1 & n:
```
csh519, line 27: I think this `if` clause executed when `n` is `[0, 2]`.
Aren't you intend just for `0`?
Please check.

credsweeper/filters/cred_card_number_check.py
csh519: Okay. I understood little more not fully..
As I see the pattern detect only 16 symbols but the code include 15 symbols or started with `00` cases.
If some card number which is 15 symbols typed like as you explained above, okay i agree.
But in `00` case.. why is it needed? for 14 symbols cases?

credsweeper/filters/cred_card_number_check.py
babenek, line 27:
```suggestion
                if 0 == 1 & n:
```

babenek: @csh519 
every second digit has to be multiplied to 2.
If the variant is more readable - let apply this.

credsweeper/filters/cred_card_number_check.py
babenek: many dumps for 64bit offset have the signature and many cases
```
0000000000000000 n wm4.cstdlib.42.38c5efcdb0f7c179bd11e611a0b94359
```
the check filters the case much quickly

credsweeper/filters/cred_card_number_check.py
kmnls, line 27:
```suggestion
                if 0 == (1 & n):  # Only for odd numbers (with 0 as a start index)
```

credsweeper/filters/cred_card_number_check.py
```diff
@@ -0,0 +1,39 @@
...
+                    x *= 2
+                    if x > 9:
+                        x -= 9
+                s += x
+
+            if 0 == s % 10:
```
kmnls, line 33: ok, the math is good

credsweeper/filters/cred_card_number_check.py
kmnls: 00 cannot start the card number but there are may core dumps with memory addresses in this format, aren't they?

kmnls: why don't you leave the check for just one 0 in the beginning? It seems enough and it may being checked in the regex.

babenek: @kmnls 
i assume, threre might be an american express card number with padding. Then left character will be ``0``.
Sequense ``40xx...`` makes a regex too complicated. e.g. [1-9]{2}[0-9]{14} - does not work.

credsweeper/filters/cred_card_number_check.py
babenek: checked. comment added

csh519 (approved):
Okay new card number rule added.

LGTM 👍

kmnls (approved):

PR #270, opened 2023-02-21
Refactoring to reduce function complexity

babenek:
## Description

The PR might be reviewed after #280 

- Refactoring recursive scan to reduce the complexity. DeepScanner was added for recursive scanning. Performance still the same. 
- UnitTests have required small changes
- source files count increased up to 100 files - fixed workflow with mypy
- dynamic analysis requires updating current seeds it will be applied in #278 

Fuzzing step wil pass after #278 

## How has this been tested?

- [x] UnitTest
- [x] FuzzTest
- [x] Benchmark


credsweeper/app.py
```diff
@@ -277,31 +278,35 @@ def file_scan(self, content_provider: ContentProvider) -> List[Candidate]:
                         data = diff
                 else:
                     logger.warning(f"Content provider {type(content_provider)} does not support deep scan")
+
                 if data:
+                    new_size_limit = self.config.size_limit if self.config.size_limit else RECURSIVE_SCAN_LIMITATION
+                    new_size_limit -= len(data)
                     data_provider = DataContentProvider(data=data,
```
csh519, line 285: What about make a new class structure regarding scan?
I think create scan directory for those scan methods and put all of scanning methods not only extract methods like below. (ex. `__zip_scan()`, `__tar_scan()`, and so on..)

csh519 (commented):
Sorry for late review.

Thank you for refactoring and i have some comments about this.

First, I want to reduce amount of `app.py`, it is too huge now.
Actually `app.py` has so many methods on it and it seems scan methods can be extracted from `app.py`.

credsweeper/app.py
babenek: Well, with the new structure i added additionaly optional pathlib.Path and ioBytesIO for our providers to use them in memory only.

credsweeper/credentials/augment_candidates.py
```diff
@@ -0,0 +1,16 @@
...
+    """Augment candidates with new_candidates if value of line data is not present in the candidates"""
+    if not new_candidates:
+        return
+    found_values = set(line_data.value for candidate in candidates  #
+                       for line_data in candidate.line_data_list)
+    for new_candidate in new_candidates:
+        for line_data in new_candidate.line_data_list:
+            if line_data.value not in found_values:
+                candidates.append(new_candidate)
+                break
```
csh519, line 16: I think it would be good to add `return` line in `augment_candidates()`.
For good readability like below.

``` python
...

candidates.extend(augment_candidates(new_candidates))

...
```

babenek: ``augment_candidates`` requires a list which has to be augmented with new candidates.
If a new candidate has the same value - it means "upper" layer of scanner (plain text usually) already found the credential.
The function reduces multiple duplicates of credential. E.g. GitHub token may be found in text which is a YAML or Json. Then, deep scan finds the same value when markup language is parsed to structure.
Only the value was chosen as key due line number and file name might be not defined in the scan.

csh519: Yup, I know about the method.
I just worried about the style of method and I think it can be reduce readability which method without return value.
Anyway if you think it is good, its okay for me too.

csh519 (approved):
Deep scanning parts are refactored.

LGTM 👍

credsweeper/credentials/augment_candidates.py
babenek: @csh519 
docsting was updated

csh519: Thank you for update.

csh519 (approved):

iuriimet (approved):

PR #279, opened 2023-03-01
Apply pathlib for path in ContentProvider

babenek:
## Description

- Add pathlib.Path to constructor of FilesProvider
- Added tests for files and bad tar file

## How has this been tested?

- [x] UnitTest


credsweeper/file_handler/text_provider.py
```diff
@@ -59,10 +69,14 @@ def get_scannable_files(self, config: Config) -> Union[List[DiffContentProvider]
...
 
         """
-        file_paths = []
+        file_paths: List[Union[str, Path]] = []
         for path in self.paths:
-            new_files = FilePathExtractor.get_file_paths(config, path)
-            if self.skip_ignored:
-                new_files = FilePathExtractor.apply_gitignore(new_files)
-            file_paths.extend(new_files)
+            if isinstance(path, str) or isinstance(path, Path):
```
csh519, line 74: This `if-else` chain is repeated here and above code(`file_path_extractor.py`).
I think it can be extracted as a method of `Util` class or other class is okay.
Please check.

babenek: OK. I will refactory the cases `` str and Path`` in #280 or #270 to avoid merge conflicts.

babenek: @csh519 
there is a doubt...
MyPy checker cannot decide whether it correct annotation. Unfortunately, elegant solution was introduced in Python 3.10 only. https://github.com/babenek/CredSweeper/pull/25 - my test
Otherwise, mypy errors might be suppressed with force, but it is not "clean code"
There are only 3 cases of the checks - i suppose, it is the maximum of  usage. If more - lets do the refactoring

PS: lambda is not recognised too
``is_likewise_path = lambda path: isinstance(path, str) or isinstance(path, Path)```

kmnls (approved):

credsweeper/file_handler/text_provider.py
csh519: Okay. As following the messanger I agreed with you.

csh519 (approved):
`pathlib` is adopted.

LGTM 👍

PR #292, opened 2023-05-01
Applied pathlib.Path where possibly

babenek:
## Description

- Applied pathlib.Path where possibly
- Simple docstrings were added to clean the warnings
- updated tests for using self.assertXXX tool from unittest
- added parametrized test for application - should help for rules manipulation

## How has this been tested?

- [x]  UnitTest


.github/workflows/test.yml
```diff
@@ -41,25 +41,30 @@ jobs:
...
+        pip install pytest pytest-random-order deepdiff
 
     - name: UnitTest with pytest
       run: |
-        pytest --random-order --random-order-bucket=global --ignore=docs --ignore=tests/test_app.py tests/
-
-    - name: ApplicationTest with pytest
-      run: |
-        pytest --random-order --random-order-bucket=global --ignore=docs tests/test_app.py
+        # put the command in one line to use in all os - various line carry style
```
kmnls, line 66:
```suggestion
        # put the command into one line to use in various OS to avoid processing differences in new line char sequence
```

credsweeper/filters/group/group.py
```diff
@@ -11,6 +11,7 @@
 
 
 class Group(ABC):
+    """Anstract Group class"""
```
kmnls, line 14:
```suggestion
    """Abstract Group class"""
```

kmnls (dismissed):

.github/workflows/benchmark.yml
```diff
@@ -110,15 +110,17 @@ jobs:
...
     - name: Upload artifact
+      if: always()
       uses: actions/upload-artifact@v3
       with:
         name: report
-        path: output.json
+        path: report.json
 
     - name: Upload artifact
+      if: always()
```
csh519, line 123: Why this field is needed?
Doesn't it run every time without this field?

credsweeper/app_path.py
```diff
@@ -0,0 +1,4 @@
+# use separate file to avoid circular import issue
+from pathlib import Path
+""" Directory of credsweeper sources """
+APP_PATH = Path(__file__).resolve().parent
```
csh519, line 4: What about move `APP_PATH` to `credsweeper/app.py`?
I think it seems more proper

``` python
...
from credsweeper.app import CredSweeper, APP_PATH
...
```

.github/workflows/benchmark.yml
babenek: ``cp $(find . -type f -name "output.json") .`` generated error when output file is the same.
Hope, it will not happen but the steps perform to upload artifacts if they exists.
They might be valuable.

credsweeper/app_path.py
babenek: Unfortunately it goes to circular dependency error.

csh519: I see. Please check this later.

.github/workflows/benchmark.yml
csh519: Yeah I see. But what relation between `always()` and checking if they exists?

babenek: it is skipped when 'cp' fails.

babenek:
@csh519 I removed your test_doc - test_data is used instead.
You can compare the produced reports.

.github/workflows/benchmark.yml
csh519: Thank you for explanation.

tests/test_main.py
```diff
@@ -681,30 +685,91 @@ def test_credit_card_number_n(self) -> None:
...
+                cfg["json_filename"] = str(tmp_file)
+                cred_sweeper = CredSweeper(**cfg)
+                cred_sweeper.run(content_provider=content_provider)
+                with open(tmp_file, "r") as f:
+                    test_result = json.load(f)
+                prepare(test_result)
+
+                diff = deepdiff.DeepDiff(test_result, expected_result)
+                if diff:
+                    print(json.dumps(test_result, ), flush=True)
```
csh519, line 772: Is it for debug?
If true, please remove this line.

babenek: It helps to find failure.
When diff present - full output is printed. Then it can be compared or replaced present test data.
Probably, output should write some text what to do...

babenek: e.g. deepdiff produces this:
```
E               + {'values_changed': {"root[18]['line_data_list'][0]['path']": {'new_value': 'teswts/samples/test.html',
E               +                                                               'old_value': 'tests/samples/test.html'}}} : {'json_filename': '/tmp/tmp81p2kcwc/doc.json', 'doc': True}
```
It is hard to understand what it means and where is the diff.
But if copy and paste print's output - git shows beautiful diff.

csh519 (approved):
Paths are changed for clean code and tests are improved.

LGTM :+1:

kmnls (approved):

PR #299, opened 2023-05-13
Performance benchmark

babenek:
## Description

- added performance test against released version
- ``--banner`` might be passed as single argument

## How has this been tested?

- [x] benchmark workflow
- [x] 10% - significant level


.github/workflows/benchmark.yml
```diff
@@ -146,3 +146,113 @@ jobs:
...
+
+    - name: Update PIP
+      run: python -m pip install --upgrade pip
+
+    - name: Install released CredSweeper
+      run: python -m pip install credsweeper
+
+    - name: Run performance benchmark
+      run: |
+        credsweeper --path requirements.txt --banner
```
csh519, line 195: Why is it needed? `requirements.txt` file scanning?

babenek: It's time to do --banner self-sufficient!

csh519: Thank you for fix!

csh519 (approved):
Checking performance of benchmark data is needed.
And adding `--banner` option as a standalone also good.

LGTM 👍

kmnls (approved):

PR #303, opened 2023-05-13
Skip binary files to scan with LATIN_1 decoding

babenek:
## Description

- detect binary files when LATIN_1 is decoding a file - it may decode anything and there is a trash for binary files.

## How has this been tested?

- [x] UnitTest


credsweeper/utils/util.py
```diff
@@ -139,18 +146,27 @@ def decode_bytes(content: bytes, encodings: Tuple[str, ...] = default_encodings)
...
         lines = []
+        binary_suggest = False
+        if encodings is None:
+            encodings = AVAILABLE_ENCODINGS
         for encoding in encodings:
             try:
-                text = content.decode(encoding)
-                if content != text.encode(encoding):
+                if binary_suggest and LATIN_1 == encoding and Util.is_binary(content):
+                    # LATIN_1 may convert any data - check them for binary formats
```
kmnls, line 159:
```suggestion
                    # LATIN_1 may convert data (bytes in range 0x80:0xFF are transformed) so skip this encoding when checking binaries
```

credsweeper/utils/util.py
```diff
@@ -322,6 +339,16 @@ def is_pdf(data: bytes) -> bool:
                 return True
         return False
 
+    @staticmethod
+    def is_elf(data: Union[bytes, bytearray]) -> bool:
+        """According https://en.wikipedia.org/wiki/Executable_and_Linkable_Format - use only 5 bytes"""
```
kmnls, line 344:
```suggestion
        """Accordingly to https://en.wikipedia.org/wiki/Executable_and_Linkable_Format use only 5 bytes"""
```

kmnls (approved):

credsweeper/utils/util.py
babenek: https://github.com/babenek/CredSweeper/commit/c3526b2684030e09f8667e95bbb6a85e655e16f6 - will be applied in next pr

csh519 (approved):
Adding skipping `LATIN_1` binary file is good.

LGTM :+1:

PR #310, opened 2023-05-17
Old atlassian pat

babenek:
## Description

- Add Atlassian old PAT pattern

## How has this been tested?

- [x] UnitTest
- [x] Benchmark


credsweeper/rules/config.yaml
```diff
@@ -724,6 +724,17 @@
     - src
     - doc
 
+- name: Atlassian Old PAT token
+  severity: info
+  type: pattern
+  values:
+    - (^|[^0-9A-Za-z.$_/+-])(?P<value>[a-zA-Z0-9]{24})([^0-9A-Za-z.$_/+-]|$)
```
babenek, line 731:
```suggestion
    - (^|[^0-9A-Za-z.$_/+-])(?P<value>[a-zA-Z0-9]{24})([^=0-9A-Za-z.$_/+-]|$)
```

credsweeper/common/morpheme_checklist.txt
```diff
@@ -584,7 +582,7 @@ fire
 firm
 first
 fix
-flash
+flas
 flat
```
csh519, line 586: Could you explain why the keyword list modified?
Is there any data before and after?

babenek: 'ash' was added and it conflicts with 'flash', 'hash' ...

csh519: Okay, then why the `ash` is added?
I'm not talking about the `flash` case only, but all.

babenek: I don't remember exactly.
- it helps filter out 'hash_xxxxx...' creds - exist 'has' and 'ash'
- 'ash' is an independent word.

iuriimet (approved):

PR #312, opened 2023-05-17
Grafana patterns

babenek:
## Description

- Add Grafana access policy token pattern and sample
- Add Grafana provisioned api key pattern and sample
- update count of found credentials in benchmark - the api key is like JWT, so it just dublicate existed rule.

## How has this been tested?

- [x] UnitTest 
- [x] Benchmark


cicd/benchmark.txt
```diff
@@ -1,5 +1,5 @@
-Detected Credentials: 4449
-result_cnt : 4056, lost_cnt : 70, true_cnt : 3672, false_cnt : 314
+Detected Credentials: 4451
+result_cnt : 4058, lost_cnt : 70, true_cnt : 3672, false_cnt : 314
```
babenek, line 2:
```suggestion
result_cnt : 4058, lost_cnt : 72, true_cnt : 3672, false_cnt : 314
```

credsweeper/filters/value_grafana_check.py
```diff
@@ -0,0 +1,38 @@
...
+            True, when need to filter candidate and False if left
+
+        """
+        if not line_data.value:
+            return True
+        with contextlib.suppress(Exception):
+            if line_data.value.startswith("glc_"):
+                # Grafana Access Policy Token
+                decoded = base64.b64decode(line_data.value[4:])
+                keys = ["o", "n", "k", "m"]
```
csh519, line 28: What about check like below? For more accuracy.

```suggestion
                keys = ['"o":', '"n":', '"k":', '"m":']
```

credsweeper/filters/value_grafana_check.py
```diff
@@ -0,0 +1,38 @@
...
+            else:
+                # Grafana Provisioned API Key
+                decoded = base64.b64decode(line_data.value)
+                keys = ["n", "k", "id"]
```
csh519, line 32: Here is too.

```suggestion
                keys = ['"n":', '"k":', '"id":']
```

credsweeper/filters/value_grafana_check.py
babenek: The payload is parsed anyway. Yeys searching in dictionary is quickly.  There might be broken json. 
Additionally json allows constructions ``"k" :`` - with a space.

babenek: Then we will skip strong verification with parsing Json structure. It might takes FP cases(very low probability, but it is)

csh519: NVM I misunderstood about this.... :tired_face:

iuriimet (approved):

PR #332, opened 2023-05-17
Additional filters for token pattern

babenek:
## Description

To support future patterns:

- Add TokenBase32 filter group for token in base32
- Add StructuredToken filter group for token which might be an encoded data
- Add TokenPattern filter group for common token
- Morpheme checklist added to filter out tokens which have two morphemes from the list

## How has this been tested?

- [x] UnitTest


credsweeper/filters/value_structured_token_check.py
```diff
@@ -0,0 +1,74 @@
...
+        return True
+
+    @staticmethod
+    def check_atlassian_struct(value: str) -> bool:
+        """Returns False if value is valid for atlassian structure 'integer:bytes'"""
+        decoded = base64.b64decode(value)
+        delimeter_pos = decoded.find(b':')
+        if delimeter_pos > 0:
+            val = decoded[:delimeter_pos].decode(LATIN_1)
+            if int(val):
```
xDizzix, line 68: If the value is not an integer an exception is raised so this method will never return True

xDizzix (commented):
Please check the comments to the code

xDizzix (commented):
Please check the comments to the code

credsweeper/filters/group/structured_token.py
```diff
@@ -0,0 +1,12 @@
...
+from credsweeper.filters import ValueStructuredTokenCheck
+from credsweeper.filters.group import Group
+
+
+class StructuredToken(Group):
+    """Structured Token"""
+
+    def __init__(self, config: Config) -> None:
+        super().__init__(config, GroupType.DEFAULT)
+        self.filters.extend([ValueStructuredTokenCheck()])
```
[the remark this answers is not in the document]
babenek: Actually, GroupType.DEFAULT inits self.filters with empty list. It can be simplified.

credsweeper/filters/value_number_check.py
```diff
@@ -0,0 +1,30 @@
...
+
+        """
+        if not line_data.value:
+            return True
+        value = line_data.value.lower()
+        if value.startswith("0x") and ValueNumberCheck.HEX_VALUE_REGEX.match(value):
+            return True
+        if ValueNumberCheck.DEC_VALUE_REGEX.match(value):
+            return True
+        return False
```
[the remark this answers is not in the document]
babenek: The code splitting helps to understand how it works with coverage statistics and just append additional code without modification existed lines.

credsweeper/filters/value_structured_token_check.py
```diff
@@ -0,0 +1,74 @@
...
+            pass
+        return True
+
+    @staticmethod
+    def check_jwt_struct(value: str) -> bool:
+        """Returns False if decoded value has json structure"""
+        decoded = base64.b64decode(value)
+        if json.loads(decoded):
+            return False
+        return True
```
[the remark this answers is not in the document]
babenek: {} and [] are not valid token :) So, let it be filtered. 
Moreover, the values will not passed here due regex of encoded string does not match
```
$ echo "{}"|base64
e30K
$ echo "[]"|base64
W10K
```

credsweeper/filters/value_structured_token_check.py
babenek: above is try-catch block.

credsweeper/common/keyword_checklist.py
```diff
@@ -30,3 +31,18 @@ def keyword_set(self) -> Set[str]:
...
+    def morpheme_set(self) -> Set[str]:
+        """Get extended set with keywords.
+
+        Return:
+            Extended set of strings
+
+        """
+        return self.__morpheme_set
+
+    @cached_property
```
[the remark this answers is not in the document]
babenek: ok. i'll check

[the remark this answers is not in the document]
babenek: @cached_property speedups a bit, but in profiler the len() invocation has fallen by order of time usage

credsweeper/common/keyword_checklist.py
```diff
@@ -13,8 +13,9 @@ class KeywordChecklist:
 
     def __init__(self) -> None:
         # used suggested text read style. split() is preferred because it strips 0x0A on end the file
-        with open(self.KEYWORD_PATH, 'r') as f:
-            self.__keyword_set = set(f.read().split())
+        self.__keyword_set = set(self.KEYWORD_PATH.read_text().split())
+        # optimized separately list with morphemes - only substring which may be separated words or 4-chars ending
```
Yullia, line 17:
```suggestion
        # optimized separately list with morphemes - only substring which may be separated words or 4-chars ending
```
The list of morphemes which can be combined to form words.
The value is considered as variable if at least two of them are present.

babenek: OK. please, use properly indent, split to 2 lines (limit is 120)

credsweeper/common/keyword_checklist.py
```diff
@@ -13,8 +13,10 @@ class KeywordChecklist:
 
     def __init__(self) -> None:
         # used suggested text read style. split() is preferred because it strips 0x0A on end the file
-        with open(self.KEYWORD_PATH, 'r') as f:
-            self.__keyword_set = set(f.read().split())
+        self.__keyword_set = set(self.KEYWORD_PATH.read_text().split())
+        # The list of morphemes can be combined to form words.
```
babenek, line 17:
```suggestion
        # The list of morphemes can be combined to form words.
```

credsweeper/common/morpheme_checklist.txt
```diff
@@ -0,0 +1,1326 @@
+000
+111
+222
+333
+444
```
csh519, line 5: Could you explain this `morpheme_checklist.txt` file came from?

credsweeper/filters/value_token_base32_check.py
```diff
@@ -0,0 +1,24 @@
...
+
+        Return:
+            True, if need to filter candidate and False if left
+
+        """
+        if not line_data.value:
+            return True
+
+        stats = PasswordStats(line_data.value)
+        return bool(0.7 > stats.strength())
```
csh519, line 24: So it means `password_strength` measure the strength with in 0 ~ 1.0?
Then why you set pass condition as `0.7`? Could you explain it more?

tests/common/test_keyword_checklist.py
```diff
@@ -31,3 +38,41 @@ def test_keyword_set_n(self):
...
+                        if i in optimized_morpheme_list[m]:
+                            # should be removed
+                            to_remove.add(optimized_morpheme_list[m])
+                        m += 1
+                else:
+                    # wrong - must be removed
+                    to_remove.add(i)
+            for x in to_remove:
+                optimized_morpheme_list.remove(x)
+            optimized = 0 == len(to_remove)
```
csh519, line 70: What about change variables like below?
They are hard to understand at once. :dizzy_face: 

```suggestion
            to_remove_set = set()
            optimized_morpheme_list_len = len(optimized_morpheme_list)
            idx = 0
            for optimized_morpheme in optimized_morpheme_list:
                idx += 1
                if optimized_morpheme.lower() == optimized_morpheme and 3 <= len(optimized_morpheme):
                    # search in rest list whether the items have the substring
                    local_idx = idx
                    while local_idx < optimized_morpheme_list_len:
                        if optimized_morpheme in optimized_morpheme_list[local_idx]:
                            # should be removed
                            to_remove.add(optimized_morpheme_list[local_idx])
                        local_idx += 1
                else:
                    # wrong - must be removed
                    to_remove_set.add(optimized_morpheme)
            for to_remove in to_remove_set:
                optimized_morpheme_list.remove(to_remove)
```

babenek: done

credsweeper/filters/value_token_base32_check.py
babenek: Tokens looked like BASE32 encoded data have only 16 characters length. 
An entropy will be weak anyway.
The "strength" calculates something like "data complexity". 
0.7 was chosen manually with false positive cases review.

credsweeper/common/morpheme_checklist.txt
babenek: The list was created manually after false positive cases review.
The approach is: 
- got a value which looks like a token "PKGVULNERABILITY"
- split long word to morphemes on English words which might be present in text "VULN" (a subset of vulnerable) and "ABILITY" (the word and postfix). 3 letter words may be acceptable if the sequence is a word.
- add the 2 morphemes to list
- run test to be sure it does not removed many others
- run scan and research next FP.
- 3 digits were found too. In real token the sequence has low probability, so i added it too.

tests/common/test_keyword_checklist.py
csh519: Thank you for a quick fix.
I've checked this with 4b28845.

credsweeper/common/morpheme_checklist.txt
csh519: Thank you for explanation, next time it would be good to add rules that using those filters.

credsweeper/filters/value_token_base32_check.py
csh519: As I told, please make the number to const.

csh519 (approved):
The filters will be used are added.

LGTM :+1:

iuriimet (approved):

PR #345, opened 2023-05-31
Rule refactoring

babenek:
## Description

- Rule refactoring to make some parameters default and pass various set of filters. When 
``  filter_type: GeneralPattern `` - string it is a group
```
  filter_type:
    - ValueCoupleKeywordCheck
    - ValuePatternCheck
```
 list with specific set of filters
- Config is passed to constructor for each rule (not necessary for all)

## How has this been tested?

- [x] UnitTest


credsweeper/rules/rule.py
```diff
@@ -37,72 +42,91 @@ class Rule:
...
+            return []
+        elif isinstance(filter_type, str):
+            # group applying
             filter_group = getattr(group, filter_type, None)
-            if filter_group is None:
-                raise ValueError(f'Malformed rule config file. Rule filter_type "{filter_type}" is invalid.')
-            self.__filters = filter_group(self.config).filters
+            if isinstance(filter_group, type) and issubclass(filter_group, Group):
+                return filter_group(self.config).filters  # type: ignore
+        elif isinstance(filter_type, list):
```
csh519, line 116: IMO `filter_type` can be changed to `Optional[None]` parameter.
For now it needs duplicated if-else chain.
What do you think?

credsweeper/rules/rule.py
```diff
@@ -204,19 +210,27 @@ def validations(self, validation_names: List[str]) -> None:
...
-        selected_validations = []
 
-        if validation_names is not None:
+        if validation_names == "" or validation_names is None:
+            return []
+        if isinstance(validation_names, str):
+            if validation_template := getattr(validations, validation_names, None):
+                return [validation_template]
+        if isinstance(validation_names, list):
+            selected_validations = []
```
csh519, line 220: This code block is same as above.
Can be changed with `Optional[list]` type.

credsweeper/rules/rule.py
```diff
@@ -37,72 +42,91 @@ class Rule:
...
+    @cached_property
     def filters(self) -> List[Filter]:
         """filters getter"""
         return self.__filters
 
-    @filters.setter
-    def filters(self, filter_type: str) -> None:
+    def _get_filters(self, filter_type: Union[None, str, List]) -> List[Filter]:
         """filters setter"""
         if filter_type == "" or filter_type is None:
```
kmnls, line 109: if not filter_type ???

credsweeper/rules/rule.py
kmnls, line 116: in the function definition there is a 'List', here it is the 'list'. Please make a decision which type you use

credsweeper/rules/rule.py
babenek: yaml.load("demo:\n  -\n", Loader=yaml.FullLoader)
{'demo': [None]}
bool(yaml.load("demo:\n  -\n", Loader=yaml.FullLoader))
True

credsweeper/rules/rule.py
babenek: ``List = _alias(list, 1, inst=False, name='List')``
alias - does not matter. In definitiion is not possibly for python 3.8 describe list of what

credsweeper/rules/rule.py:208: in Rule
    def _get_validations(self, validation_names: Union[None, str, list[str]]) -> List[Validation]:
E   TypeError: 'type' object is not subscriptable

credsweeper/rules/rule.py
```diff
@@ -37,72 +42,95 @@ class Rule:
...
-            self.__filters = []
-        else:
+    def _get_filters(self, filter_type: Union[None, str, List[str]]) -> List[Filter]:
+        """
+            filter_type: str - applies Group of filter
+                         list - creates specific set of Filters
+        """
+        if not filter_type:
+            # empty line or skipped element mean empty list of filters
+            return []
```
csh519, line 114: It seems this `if` condition can be removed.

There are 3 cases for this `if` condition.
1. `filter_type` = `None`
  - It can be handled by `raise ValueError` which in tail of function.
2. `filter_type` = `""`
  - It can be handled by `raise ValueError` which in tail of function, through the `elif isinstance(filter_type, str):` if-else chain.
3. `filter_type` = `[]`
  - It can be handled on `elif isinstance(filter_type, list):` if-else chain. It will return empty list `[]`.

babenek: Then, it means filter_type will be mandatory  field. Some tests failed due were developed without the check.

csh519: Yeah that's right. Thank you for a quick fix.

csh519 (approved):
The classes related Rule are refactored.

LGTM :+1:

iuriimet (approved):

PR #346, opened 2023-06-05
reduced html is parsed too with additional line ending for <p>

babenek:
## Description

- reduced html text without <header>, <body> might be parsed too
- add line end for <p> tags to split values

## How has this been tested?

- [x] UnitTest


csh519 (approved):
Improving html parsing logic.

LGTM :+1:

csh519 (commented):
BTW, what about just use html parser which is opensource?

PR #349, opened 2023-06-12
Fix wrong rules and samples

csh519:
## Description

There were some rules and samples that couldn't working properly.
And some rules have never been detected at all.
So I changed some rules to `use_ml: false` and fixed samples.

- Fix Rules and Samples

## How has this been tested?

- [X] Local tested


babenek (approved):
Ok.
The samples look more close to real credentials.

Yullia (approved):

PR #357, opened 2023-06-14
min_line_len is mandatory field

babenek:
## Description

- min_line_len is mandatory field because without them a rule does not work

## How has this been tested?

- [x] UnitTest


xDizzix (approved):
+

csh519 (approved):
Thank you for move `MIN_LINE_LEN` to mandatory attribute.

LGTM :+1:

PR #367, opened 2023-06-23
apply ValuePemPatternCheck for PEM rule

babenek:
## Description

REVIEW AFTER #370 !!!

- Applied forgotten ValuePemPatternCheck for PEM scanner
- fix offset limit for the scanner
- refactored PEM detector
- morphemes list updatet with VoC
- set 2000 as maximal line length

## How has this been tested?

- [x] UnitTest
- [x] Test Benchmark


credsweeper/scanner/scan_type/pem_key_pattern.py
```diff
@@ -38,35 +40,39 @@ def run(cls, config: Config, rule: Rule, target: AnalysisTarget) -> Optional[Can
...
             Boolean. True if PEM key, False otherwise
 
         """
+        limit_line = start_line + 190
         lines = cls.strip_lines(lines)
         lines = cls.remove_leading_config_lines(lines)
         key_data = ""
         for line_num, line in enumerate(lines):
-            if line_num >= 190:
+            if line_num >= limit_line:
```
csh519, line 68: Why the `start_line` parameter is needed?
Comparing `line_num` is just compare not to over `190` lines, not the original line number(`start_line` + `190`).
Please check.

credsweeper/scanner/scan_type/pem_key_pattern.py
```diff
@@ -38,35 +40,39 @@ def run(cls, config: Config, rule: Rule, target: AnalysisTarget) -> Optional[Can
         """
         assert rule.pattern_type == rule.PEM_KEY_PATTERN, \
             "Rules provided to PemKeyPattern.run should have pattern_type equal to PEM_KEY_PATTERN"
-
-        if cls.is_pem_key(target.lines[target.line_num:], config):
+        global pem_pattern_check
+        if not pem_pattern_check:
+            pem_pattern_check = ValuePemPatternCheck(config)
```
csh519, line 45: IMO `pem_pattern_check` can be defined as `cls.pem_pattern_check`.
Just define `cls.pem_pattern_check` variable as member.
``` python
class PemKeyPattern(ScanType):
        // ...
        cls.pem_pattern_check = ValuePemPatternCheck(config)
        // ...
       
        def run():
                // ...
```

babenek: Thank you for the suggestion! Fix looks better.

credsweeper/scanner/scan_type/pem_key_pattern.py
babenek: Refactored with limit of 190 lines before to skip extra copies.

credsweeper/scanner/scan_type/pem_key_pattern.py
```diff
@@ -58,15 +60,12 @@ def is_pem_key(cls, lines: List[str], config: Config) -> bool:
         lines = cls.strip_lines(lines)
         lines = cls.remove_leading_config_lines(lines)
```
csh519, line 61: About checking line length is over `190` or not, this method `remove_leading_config_lines()` can reduce size of `lines`.
So isn't it good to just remain line length check code here?

Like below.

credsweeper/scanner/scan_type/pem_key_pattern.py
```diff
@@ -58,15 +60,12 @@ def is_pem_key(cls, lines: List[str], config: Config) -> bool:
...
         key_data = ""
-        for line_num, line in enumerate(lines):
-            if line_num >= 190:
-                return False
```
csh519, line 63:
``` python
if len(lines) >= 190:
    return False
```

credsweeper/scanner/scan_type/pem_key_pattern.py
csh519: Thank you for a quick fix!
I've checked this change with cb13007.

credsweeper/scanner/scan_type/pem_key_pattern.py
csh519: I commented few.. please check below.

credsweeper/scanner/scan_type/pem_key_pattern.py
babenek: Opposite. It will skip huge files. 
```remove_leading_config_lines`` does not trim ending so there might be a long tail over 190 lines with remaining.
Additionally, enumerate does additional complexity. 
If the key will be not found in given 190 lines - it returns False at end of function.
Probably there should be a huge refactoring...

PR #376, opened 2023-07-07
Add lower_path to check_exclude_file() for exclude exact file path

csh519:
## Description

`FilePathExtractor.check_exclude_file()` method has a bug before.
The method just convert a `path` to lower case, so when checking the exclude file path, it didn't work properly.

I fixed this bug with adding `lower_path` parameter to `check_exclude_file()`.
`check_exclude_file()` uses `lower_path` on pattern matching and uses `path` on excluding exact paths and extensions.

- Fix bug of `check_exclude_file()`

## How has this been tested?

- [X] Local tested

### DIFF
--- a/credsweeper/file_handler/file_path_extractor.py
+++ b/credsweeper/file_handler/file_path_extractor.py
@@ -137,7 +137,7 @@ class FilePathExtractor:
         for exclude_path in config.exclude_paths:
             if exclude_path in path:
                 return True
-        file_extension = Util.get_extension(path, lower=False)
+        file_extension = Util.get_extension(path)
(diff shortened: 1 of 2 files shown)


babenek (commented):
May i propose https://github.com/csh519/CredSweeper/pull/6 to avoid wrong behaviour in future?

csh519:
> May i propose [csh519#6](https://github.com/csh519/CredSweeper/pull/6) to avoid wrong behaviour in future?

@babenek 
I approved and merged your PR. Thank you for the details.

babenek (approved):
Good. The bug must not appear again.

iuriimet (approved):

PR #377, opened 2023-07-10
Refuzzed with input size 2048

babenek:
## Description

- Fuzzing seeds were updated to use input size 2048 bytes with cover extra line width case. Additionally 512 bytes are used for mocking API validation.
- increased fuzzing coverage up to 83%

## How has this been tested?

- [x] FuzzTest only


kmnls (approved):
Agree

csh519 (commented):
Is `auxilary.py` is needed to merge?
It seems just write logs to file.

babenek:
> 

The file is helpful to generate mock response for api validation in case when samples were modified. E.g. size is changing. 
The mocking responses may be hardcoded, but sometime fuzzer was finding something interesting when modified the part of seed.
The tool is for support fuzzing only.

iuriimet (approved):

PR #380, opened 2023-07-16
Memory consumption optimization

babenek:
## Description

- Refactoring for memory consumption reducing - applied generators.
- Removed unused diff types  ( ADDED_ACCOMPANY,    DELETED_ACCOMPANY) because they never pass from current whatthepath library
- with huge synthetic data results:
Release version:
```
Maximum resident set size (kbytes): 3520084
```
This version:
```
Maximum resident set size (kbytes): 1605756
```

## How has this been tested?

- [x] UnitTest 
- [x] Test Benchmark - no changes in scores
- [x] Fuzzing


credsweeper/credentials/line_data.py
```diff
@@ -32,13 +32,15 @@ def __init__(
             self,  #
             config: Config,  #
             line: str,  #
+            line_pos: int,  #
```
csh519, line 35: Could you please explain more about `line_pos`?
As I read the code it is very similar with `line_num`.

credsweeper/scanner/scanner.py
```diff
@@ -31,17 +33,16 @@ class Scanner:
...
-        self.__scanner_for_rule: Dict[str, Type[ScanType]] = {}
-        self.rules: List[Rule] = []
         # init with MAX_LINE_LENGTH before _set_rules
         self.min_keyword_len = MAX_LINE_LENGTH
         self.min_pattern_len = MAX_LINE_LENGTH
         self.min_pem_key_len = MAX_LINE_LENGTH
-        self._set_rules(rule_path)
+        self.rules_scanners: List[Tuple[Rule, Type[ScanType]]] = []
+        self._set_rules_scanners(rule_path)
         self.min_len = min(self.min_pattern_len, self.min_keyword_len, self.min_pem_key_len,
```
csh519, line 42: IMO `min_len` property would be good to move to `ContentProvider` class.
Scanner doesn't use `min_len` directly and it just pass the `min_len` to other methods.

credsweeper/credentials/line_data.py
babenek: line_pos is used in the class for support MultiPattern
old: https://github.com/Samsung/CredSweeper/blob/3989a25b8c40527aece9c63a54008070ef39d7a3/credsweeper/scanner/scan_type/multi_pattern.py#L45
- it may be wrong when e.g. xml was parsed with "sieve" numeration. 
Additionally, diff parsing has to used emply lines to support "line_num" which is an information only. It taked extra memory too.

credsweeper/scanner/scanner.py
babenek: Yes, it might speedup for many providers and increase code efficiency.
But there might be an issue when a Provider is created outside (and/or before) CredSweeper an it does not know which minimal line length is acceptable for Scanner.
e.g. https://github.com/Samsung/CredSweeper/blob/3989a25b8c40527aece9c63a54008070ef39d7a3/credsweeper/app.py#L224

csh519: I see.
Then what about set `min_len` as a property of `ContentProvider` class?
It can be added on `file_scan()` method of `CredSweeper` class.

https://github.com/Samsung/CredSweeper/blob/3989a25b8c40527aece9c63a54008070ef39d7a3/credsweeper/app.py#L286-L316

Maybe like below.

``` python
    def file_scan(self, content_provider: Union[DiffContentProvider, TextContentProvider]) -> List[Candidate]:
        """Run scanning of file from 'file_provider'.

        Args:
            content_provider: content provider object to scan

        Return:
            list of credential candidates from scanned file

        """
        candidates: List[Candidate] = []
        content_provider.min_len = self.scanner.min_len # <-------- min_len assignment.
        logger.debug("Start scan file: %s %s", content_provider.file_path, content_provider.info)

```

babenek: The changes: https://github.com/Samsung/CredSweeper/pull/384/commits/48b13cefaf0eabd88077e5be8bec943ae9916f54
It looks not user friendly, when ``file_scan`` changes something in user data.
Performance benchmark scores have no valuable difference.

csh519: Thank you for testing with kindly changes.
Okay let's go with now changes.

credsweeper/credentials/line_data.py
csh519: I understood about it and concerns.
Thank you for explanation.

csh519 (approved):
Thank you for optimize memory usage and adding `line_pos` for more accurate scanning.

LGTM :+1:

xDizzix (approved):
Memory consumption reduced

PR #382, opened 2023-07-18
Slight changes

babenek:
## Description

- system resources measurement for benchmark workflow
- value_lower for line_data to reduce lower() invocation
- some light changes to reduce next PR complexity

## How has this been tested?

- [x] UnitTest
- [x] Benchmark performance

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -200,7 +200,6 @@ jobs:
         FINISH_TIME=$(date +%s)
         RELEASE_TIME=$(( ${FINISH_TIME} - ${START_TIME} ))
         if [ 0 -lt ${RELEASE_TIME} ]; then
-          echo ${RELEASE_TIME}>stopwatch_${{ matrix.code-type }}_${{ matrix.python-version }}_result.txt
           echo Elapsed $(date -ud "@${RELEASE_TIME}" +"%H:%M:%S")
         else
           echo "Wrong result '${RELEASE_TIME}'"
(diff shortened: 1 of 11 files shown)


credsweeper/credentials/candidate.py
```diff
@@ -32,78 +32,20 @@ def __init__(self,
                  config: Config,
                  validations: List[Validation] = None,
                  use_ml: bool = False) -> None:
+        self.line_data_list = line_data_list
+        self.patterns = patterns
+        self.rule_name = rule_name
+        self.severity = severity
+        self.config = config
+        self.validations: List[Validation] = validations if validations is not None else []
+        self.use_ml = use_ml
```
csh519, line 41: It seems all type hints are removed.. Is it helpful on memory reduction?

csh519 (commented):
First of all.. I wonder is it really needed to reduce the memory usage of CredSweeper now..
IMO old structure looks better to maintain.
Could you explain why this PR is needed?

credsweeper/credentials/candidate.py
babenek: The hints are not necessary here - they are passed from function and MyPy recognizes it. Just code cleanup. 
The hints do not impact on memory reducing.

.github/workflows/benchmark.yml
```diff
@@ -184,6 +184,10 @@ jobs:
       with:
         python-version: ${{ matrix.python-version }}
 
+    - name: Add synthetic huge data
```
babenek, line 187: @csh519 
huge file with useless AnalysisTarget requires too much memory. It happens in some repos (by VoC)

csh519: Yup. Looks good.

credsweeper/filters/value_not_part_encoded_check.py
```diff
@@ -33,7 +33,7 @@ def run(self, line_data: LineData, target: AnalysisTarget) -> bool:
             return True
 
         if line_data.line_num == target.line_num \
-                and line_data.line_len == target.line_len \
+                and len(line_data.line) == target.line_len \
```
csh519, line 36: It seems this change is for reducing memory usage.. but as a trade off it seems it can decrease the performance..
Did you compared them?

credsweeper/credentials/candidate.py
csh519: Okay. Let's remove type hints.

credsweeper/filters/value_not_part_encoded_check.py
babenek: The filter is invoked to rarely compare with AnalysisTarget and line_data.line_len is used only once - no sense for cached property.
Measurement shows less than 1% diff in performance.

csh519: Okay. Thank you for the details. :+1:

csh519 (approved):
Thank you for the first step of memory reduction.

LGTM :+1:

iuriimet (approved):

PR #383, opened 2023-07-18
Code cleaning and refuzzing

babenek:
## Description

- Clean unused code
- Update HTML deep scan to avoid long scan after fuzzing.
- Enable ``required_regex`` performace speedup (10%)
- Re-fuzzed seeds and gain 84% coverage

## How has this been tested?

- [x] UnitTest 
- [x] Test Benchmark


credsweeper/rules/config.yaml
```diff
@@ -6,6 +6,7 @@
   filter_type:
     - CreditCardNumberCheck
   min_line_len: 16
+  required_regex: ([a-zA-Z0-9_/+-]{15,})
```
csh519, line 9: Checking the line has more than 15 characters is proved to increase the performance of `required_regex`?
I asked this, because the performance might be increased by code block of `Scanner` class.
As you can see checking regex function is changed with `search()`.
https://github.com/Samsung/CredSweeper/blob/235ccce812a69598a9216ff5bdcbf22cfbc17d83/credsweeper/scanner/scanner.py#L139-L146

credsweeper/scanner/scanner.py
```diff
@@ -91,21 +91,6 @@ def yield_rule_scanner(
...
-            if substring in line:
-                return False
-        return True
-
-    @staticmethod
-    def _required_regex_not_matched(required_regex: re.Pattern, line: str) -> bool:
-        """ returns True if line does not matched required_regex """
-        if required_regex.match(line):
-            return False
-        return True
```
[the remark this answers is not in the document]
babenek: The methods were not used. 
If `` _required_substrings_not_present`` is invoced in  scan in inner loop - it slowdowns ~6%
``_required_regex_not_matched`` invocation slowndowns too. 
Moreover, matched_regex map is used locally and decreases performance if passed to a method.

credsweeper/rules/config.yaml
babenek: ``search`` was replaced to ``match``  https://github.com/Samsung/CredSweeper/pull/388/commits/d3dd3653cd50dcfbe1ab36a7d301be2615614192
locally speedup is 2%

credsweeper/scanner/scanner.py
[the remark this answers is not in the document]
csh519: Okay.
It would be good to improve the performance.. but it just make the code maintenance a little bit harder.

credsweeper/rules/config.yaml
csh519: I see, thank you for explanation.

csh519 (approved):
Thank you for cleaning the code and updating fuzz corpus.

LGTM :+1:

iuriimet (approved):

PR #390, opened 2023-07-30
NDJSON is parsing with "-- depth" option

babenek:
## Description

- Add [NDJSON](http://ndjson.org/) parse structure with --depth option. It performs parse long lines which are skipped by size.

## How has this been tested?

- [x] UnitTest


credsweeper/app.py
```diff
@@ -156,6 +156,9 @@ def _use_ml_validation(self) -> bool:
         if not self.credential_manager.candidates:
             logger.info("Skipping ML validation due to no candidates found")
             return False
+        if not any(x.use_ml for x in self.credential_manager.candidates):
+            logger.info("Skipping ML validation due not applicable")
+            return False
```
csh519, line 161: It seems this `if` clause is wrong.
If there is at least one case where ML is not used for `credential_manager`, this change skips ML Validation.

And we have already skipping logic on `post_processing()`, please check below.
https://github.com/Samsung/CredSweeper/blob/2401c3dcde3f9e9f5e2648e3b5fa61b9edd58a16/credsweeper/app.py#L335-L337

babenek: Thank you for review. I refactored the skipping. Thus we have to group candidates to decide whether ml_validator must be created or not.

csh519 (approved):
Thank you for adding `NDJSON` parsing logic and other improvements.

LGTM :+1:

credsweeper/app.py
csh519: Yeah I checked too.
We already have groups and we can decide whether creating `ml_validator` or not by the groups.
I checked all changes are reflected well with f0108fa.

Thank you.

xDizzix (approved):
+

PR #397, opened 2023-08-05
Refactored rule type - separated "multi" type

babenek:
## Description

- Separated old multi_pattern to dedicated "multi"
- fix unused "else" in loop

## How has this been tested?

- [x] UnitTest
- [x] Test Benchmark

### DIFF
--- a/credsweeper/rules/rule.py
+++ b/credsweeper/rules/rule.py
@@ -63,14 +63,11 @@ class Rule:
             self.__rule_type: RuleType = rule_type
         else:
             self._malformed_rule_error(rule_dict, Rule.TYPE)
-        self.__patterns: List[re.Pattern] = []
-        self._init_patterns(rule_dict[Rule.VALUES])
+        self.__patterns = self._init_patterns(rule_dict[Rule.VALUES])
         # auxiliary fields
-        self.__filters: List[Filter] = []
-        self._init_filters(rule_dict.get(Rule.FILTER_TYPE))
+        self.__filters = self._init_filters(rule_dict.get(Rule.FILTER_TYPE))
         self.__use_ml = bool(rule_dict.get(Rule.USE_ML))
-        self.__validations: List[Validation] = []
-        self._init_validations(rule_dict.get(Rule.VALIDATIONS))
+        self.__validations = self._init_validations(rule_dict.get(Rule.VALIDATIONS))
         self.__required_substrings = [i.strip().lower() for i in rule_dict.get(Rule.REQUIRED_SUBSTRINGS, [])]
         self.__has_required_substrings = bool(self.__required_substrings)
         required_regex = rule_dict.get(Rule.REQUIRED_REGEX)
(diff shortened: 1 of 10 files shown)


Yullia (approved):
Approved

credsweeper/rules/rule.py
```diff
@@ -69,14 +63,14 @@ def __init__(self, config: Config, rule_dict: Dict) -> None:
             self.__rule_type: RuleType = rule_type
         else:
             self._malformed_rule_error(rule_dict, Rule.TYPE)
-        self.__patterns = Rule._get_patterns(self.rule_type, rule_dict[Rule.VALUES])
+        self.__patterns: List[re.Pattern] = []
+        self._init_patterns(rule_dict[Rule.VALUES])
```
csh519, line 67: What about move `self.__patterns` to `_init_patterns()`?
It can make reduce the lines of `__init__()`.

credsweeper/rules/rule.py
```diff
@@ -69,14 +63,14 @@ def __init__(self, config: Config, rule_dict: Dict) -> None:
...
         # auxiliary fields
-        self.__filters = self._get_filters(rule_dict.get(Rule.FILTER_TYPE))
-        self.__pattern_type = Rule._get_pattern_type(self.rule_type, len(self.patterns))
-        if 2 < len(self.__patterns):
-            logger.warning(f"Rule {self.rule_name} has extra patterns. Only two patterns supported.")
+        self.__filters: List[Filter] = []
+        self._init_filters(rule_dict.get(Rule.FILTER_TYPE))
```
csh519, line 70: Same here as above :)

credsweeper/rules/rule.py
```diff
@@ -69,14 +63,14 @@ def __init__(self, config: Config, rule_dict: Dict) -> None:
...
         self.__use_ml = bool(rule_dict.get(Rule.USE_ML))
-        self.__validations = self._get_validations(rule_dict.get(Rule.VALIDATIONS))
+        self.__validations: List[Validation] = []
+        self._init_validations(rule_dict.get(Rule.VALIDATIONS))
```
csh519, line 73: And here too.

credsweeper/rules/rule.py
babenek: I scare, some of development tools are not properly analyse the code where a variable is defined (and typed) ontside __init__.
That why i keep this approach

csh519: Okay..
But as you know previous approach of initializing the variables was `get_xxx()` and it can reduce the code size and more easy to read the code.
Why you changed the methods?

babenek: Malformed rule exception and warnings might be raised with the methods. To point which rule is fail we've got  ``self.rule_name`` already. Also self.config is used.
I decided reduce complexity of arguments for the private methods - only configuration is passed.
Suppose, a return is extra for "self" method when it may be set right in the method.

csh519: Yup I know about the `self.rule_name`.
And IMO below code block is more readable and can reduce the LOC.. isn't it?

```suggestion
        self.__patterns: List[re.Pattern] = self._init_patterns(rule_dict[Rule.VALUES])
```

babenek: refactoring was done

csh519 (approved):
Thank you for adding new rule type `MULTI`.

LGTM :+1:

credsweeper/rules/rule.py
csh519: Thank you for fix! I checked.

iuriimet (approved):

PR #406, opened 2023-08-11
BUGFIX: Update search for LineData

babenek:
## Description

- add optional matched object for LineData
- FIX: Refactored searching for target when first found item is filtered (in single line only first item was found and it might be filtered)

## How has this been tested?

- [x] UnitTest
- [x]  rollback custom workflow after  https://github.com/Samsung/CredData/pull/42 merge
- [x] Benchmark


.github/workflows/benchmark.yml
```diff
@@ -62,7 +62,8 @@ jobs:
     - name: Checkout CredData
       uses: actions/checkout@v3
       with:
-        repository: Samsung/CredData
+        repository: babenek/CredData
+        ref: finditer
```
```suggestion
        repository: Samsung/CredData
```

credsweeper/common/morpheme_checklist.txt
```diff
@@ -93,6 +93,7 @@ adapt
 add
 ader
 adjust
+admi
```
csh519, line 96: I think it can make filter the line which contain `admin` keyword.. like `admin_key`, `admin_secret`, and so on..

credsweeper/deep_scanner/deep_scanner.py
```diff
@@ -191,8 +191,8 @@ def structure_scan(
                                                             file_path=struct_provider.file_path,
                                                             file_type=struct_provider.file_type,
                                                             info=f"{struct_provider.info}|STRUCT:{key}")
-                candidates.extend(self.structure_scan(val_struct_provider, depth, recursive_limit_size))
-
+                new_candidates = self.structure_scan(val_struct_provider, depth, recursive_limit_size)
+                candidates.extend(new_candidates)
```
csh519, line 195: Is there any reason to add `new_candidates` and extend it to `candidates`?

credsweeper/scanner/scan_type/scan_type.py
```diff
@@ -53,8 +53,6 @@ def filtering(cls, config: Config, target: AnalysisTarget, line_data: LineData,
             If `use_filters` option is false, always return False
 
         """
-        if not config.use_filters:
-            return False
```
csh519, line 57: Is it removed due to supporting `MultiPattern`?

credsweeper/scanner/scan_type/scan_type.py
```diff
@@ -81,32 +79,17 @@ def get_line_data(
...
-
-        """
-        if pattern.search(line):
-            return True
-        return False
+        for _match in pattern.finditer(target.line):
+            logger.debug("Valid line for pattern: %s in file: %s:%d in line: %s", pattern, target.file_path,
+                         target.line_num, target.line)
+            line_data = LineData(config, target.line, target.line_pos, target.line_num, target.file_path,
+                                 target.file_type, target.info, pattern, _match)
```
csh519, line 86: Adding `_match` to parameter of `LineData` reduces the `pattern.search()` calling count in half compared to the previous one! :eyes:

credsweeper/scanner/scanner.py
```diff
@@ -124,7 +124,7 @@ def scan(self, provider: ContentProvider) -> List[Candidate]:
             matched_pattern = target_line_stripped_len >= self.min_pattern_len
             matched_multi = target_line_stripped_len >= self.min_multi_len
 
-            if not (matched_keyword or matched_pem_key or matched_pattern):
+            if not (matched_keyword or matched_pem_key or matched_pattern or matched_multi):
```
csh519, line 127: :eyes:

credsweeper/deep_scanner/deep_scanner.py
babenek: Only easiest debug purpose in single file.

credsweeper/scanner/scan_type/scan_type.py
babenek: No. the check is moved to  ``get_line_data``.
When first candidate is filtered - second might be not filterd.

csh519: IMO `MultiPattern` should be `and` condition, not `or`.
`MultiPattern` is meaningful when both patterns are detected..

credsweeper/deep_scanner/deep_scanner.py
csh519: I see.

credsweeper/common/morpheme_checklist.txt
babenek: Yes. The list predestination is to filter out true random tokens and human typed example tokens. In true random token the sequence 'admin'  is too rarely.

credsweeper/scanner/scan_type/scan_type.py
babenek: Yes, ``MultiPattern`` requires two patterns (from values) match https://github.com/babenek/CredSweeper/blob/c9926c8ff69dd304033b0a2d8e2633db02cec6ce/credsweeper/scanner/scan_type/scan_type.py#L111
Workaround in the PR is to fix case when two different credentials are in single line. Currently Credsweeper cannot find second with one rule.

babenek: Let me show you: https://github.com/babenek/CredSweeper/commit/2a56e8117347f83c2eda39863008446c4e600c45 - dummy key hides real key.
Suppose, there might be a bit bigger refactoring ...

credsweeper/common/morpheme_checklist.txt
csh519: Ah.. I mistook it for checking `variable`..
It just check the `value` and yes, `admin` should to included in this checklist.

credsweeper/scanner/scan_type/scan_type.py
csh519: Ah.. I see.

So, If there are more than 2 detected patterns in one line, and if the first detected pattern just filtered then the others skipped also.
That's why you move the `if` condition into the loop, I understood.

csh519 (approved):
Thank you for improve the search logic of `LineData`.

LGTM :+1:

PR #407, opened 2023-08-12
Update ValueFilePathCheck filter

babenek:
## Description

- Add entropy check for the filter due base64 has '/' in the sequence and filtered out like unix path. 
- add low entropy filtering for token checks (it never happens due size of token <value> is big enough)

## How has this been tested?

- [x] UnitTest
- [x] Benchmark


credsweeper/filters/value_file_path_check.py
```diff
@@ -27,13 +29,18 @@ def run(self, line_data: LineData, target: AnalysisTarget) -> bool:
...
-        for i in " !$`&*()+":
-            if i in line_data.value:
-                contains_special_characters = True
+        value = line_data.value
+        contains_unix_separator = '/' in value and not value.endswith('=')
+        if contains_unix_separator:
+            # base64 encoded data might look like linux path
+            min_entropy = ValueEntropyBase64Check.get_min_data_entropy(len(value))
+            contains_unix_separator = (0 == min_entropy
+                                       or min_entropy > Util.get_shannon_entropy(value, Chars.BASE64STD_CHARS.value))
```
csh519, line 38: I think it would be good to there is description annotation here.
Could you please add some annotation on here?

csh519 (approved):

credsweeper/filters/value_file_path_check.py
babenek, line 38:
```suggestion
            # get minimal entropy to compare with shannon entropy of found value
            # min_entropy == 0 means that the value cannot be checked with the entropy due high variance
            contains_unix_separator = (0 == min_entropy
                                       or min_entropy > Util.get_shannon_entropy(value, Chars.BASE64STD_CHARS.value))
            # low shannon entropy points that the value may be not a high randomized value in base64
```

babenek: @csh519 please, review proposed annotation

csh519: Could you remove space of `may be` -> `maybe`?
Others are good :)

csh519 (approved):
Thank you for update the filters to check `min_entropy == 0`.

LGTM :+1:

Yullia (approved):
Approved

PR #409, opened 2023-08-23
Phone number detection

babenek:
## Description

- Add international phone format number as PII

## How has this been tested?

- [x] UnitTest
- [x] TODO: rollback custom benchmark after https://github.com/Samsung/CredData/pull/44
- [x] Benchmark


credsweeper/rules/config.yaml
```diff
@@ -1,3 +1,15 @@
+- name: Phone
+  severity: info
+  type: pattern
+  values:
+    - (^|[^0-9A-Za-z])(?P<value>\+[1-9][0-9]{6,14})([^=0-9A-Za-z]|$)
```
csh519, line 5: Could you explain it more about `{6,14}`?

As I see the page which you linked(https://en.wikipedia.org/wiki/List_of_mobile_telephone_prefixes_by_country).
The minimum length of `Size of NN` was 4 and maximum was 13.

babenek: NN- means after country prefix. e.g. 

Algeria  +213 NN= 4 -> full phone +2130000 (7 digits)
first digit in the regex is not 0 [1-9] then we found 6 digits.
max number just decide it should be 15 (maximum in https://gist.githubusercontent.com/ally-commits/9073ff23fc7f96fab1290fdec22775bc/raw/0232c21262e651cd0837105c26e2ad0546b06459/Countries%2520Phone%2520Number%2520Length)

csh519: Thank you for explanation. I understood well.

.github/workflows/benchmark.yml
```diff
@@ -62,7 +62,8 @@ jobs:
     - name: Checkout CredData
       uses: actions/checkout@v3
       with:
-        repository: Samsung/CredData
+        repository: babenek/CredData
+        ref: phone
```
```suggestion
        repository: Samsung/CredData
```

Yullia (approved):
Approved

csh519 (approved):
Thank you for adding new rule `Phone`.

LGTM :+1:

PR #410, opened 2023-08-25
Collect start-end position of value in found credential

babenek:
## Description

- Add start-end position of found values in line_data_list
- replace tuple of separator position to two separated integers to avoid possible misunderstandings during export to json (tuple is represented as list)

## How has this been tested?

- [x] UnitTest
- [x] Benchmark


credsweeper/credentials/line_data.py
```diff
@@ -49,6 +49,9 @@ def __init__(
         self.pattern: re.Pattern = pattern
         # do not store match object due it cannot be pickled with multiprocessing
 
+        # start - end position of matched object
+        self.value_start = -2
```
babenek, line 53: -2 has been chosen to distinguish with standard -1 value when a string was not found.
Anyway, the value is sub-zero and may be skipped if appears.

tests/data/output.json
```diff
@@ -1762,6 +1926,8 @@
                 "path": "tests/samples/pem_key",
                 "info": "",
                 "value": null,
+                "value_start": -2,
```
babenek, line 1929: The case, when a pattern was not found

credsweeper/credentials/line_data.py
csh519: Does -1 has other mean except string was not found?

csh519: I see.

csh519 (approved):
`value_start` and `value_end` fields are added.

LGTM :+1:

PR #412, opened 2023-08-26
Rule: multiple values search in single line

babenek:
## Description

- Extended "pattern" and "keyword" rule types to search multiple values in one line. First value pattern is primary, others - might be in various order.
- HTML representation : collect matched keywords fields only 
- simple PII info rule (name, birthday, gender keywords)

## How has this been tested?

- [x] UnitTest
- [x] TODO: rollback custom markup after https://github.com/Samsung/CredData/pull/45
- [x] Benchmark


.github/workflows/benchmark.yml
```diff
@@ -62,7 +62,8 @@ jobs:
     - name: Checkout CredData
       uses: actions/checkout@v3
       with:
-        repository: Samsung/CredData
+        repository: babenek/CredData
+        ref: pii
```
```suggestion
        repository: Samsung/CredData
```

credsweeper/rules/config.yaml
```diff
@@ -1,3 +1,19 @@
+- name: PII
+  severity: critical
```
babenek, line 2:
```suggestion
  severity: info
```

iuriimet (dismissed):

credsweeper/deep_scanner/html_scanner.py
```diff
@@ -19,7 +19,7 @@ def data_scan(
             depth: int,  #
             recursive_limit_size: int) -> List[Candidate]:
         """Tries to represent data as html text and scan as text lines"""
-        if data_provider.represent_as_html():
+        if data_provider.represent_as_html(self.keyword_check, depth, recursive_limit_size):
```
csh519, line 22: It seems unnatural passing `self.keyword_check` method here...
As I see, `represent_as_html()` method just check if the data can be parsed as html or not.
What about leave `represent_as_html()` as it is and call `keyword_check()` inside of `html_scanner.py`?

credsweeper/deep_scanner/deep_scanner.py
```diff
@@ -43,6 +44,17 @@ def __init__(self, config: Config, scanner: Scanner) -> None:
         """
         self.__config = config
         self.__scanner = scanner
+        self.__keywords = set()
+        for keyword_rule in [x[0] for x in scanner.rules_scanners if RuleType.KEYWORD == x[0].rule_type]:
+            self.__keywords.update(set(keyword_rule.required_substrings))
```
csh519, line 49: As you know `Rule` class already has `required_substrings` and above logic of updating `__keywords` set seems duplicated and not that intuitive.
Is it impossible to change checking `required_substrings` by using `Rule` class?

babenek: Probably i can rename self.__keywords to self.__all_rules_required_substrings ?
Iteration for rules requires more time than for the set.

credsweeper/deep_scanner/html_scanner.py
babenek: The represent method does not only check but perform the representation.
The keyword_check is passed to reduce code duplication. It helps also skip table headers combination when a header does not contain required substring.

credsweeper/deep_scanner/deep_scanner.py
csh519: I mean the structure of each feature(`DeepScanner`, `Rule`, and so on) and its role can be mixed..
Actually each `required_substrings` is related with its `Rule` as you know..
So it is hard to agree with merging all `required_substrings` into one variable `__keywords`.

credsweeper/deep_scanner/html_scanner.py
csh519: Hmm..
Checking `required_substrings` already existed in `Scanner` class.
https://github.com/Samsung/CredSweeper/blob/1bf1d1f64372ac9adb576e9aba1d98f9e29d3c8d/credsweeper/scanner/scanner.py#L137-L142

Is it really needed to be checked on `HTMLScanner` class?
Maybe it can increase the performance.. but I think it makes code more complicated.

credsweeper/deep_scanner/deep_scanner.py
babenek: Yes, each rule has separated required_strings list to trigger.
In html_representation we need to know pass the header to targets or not. In future the target will be scanned with rule as usually.
I'll do refactoring ...

csh519: As you told, `represent_as_html()` method's role is represent the data to html not to check keywords and ignore or not.
After representation, `Rule` class will be used to filter the target is valid or not.
Please mind single responsibility principle.

credsweeper/deep_scanner/deep_scanner.py
```diff
@@ -1,7 +1,8 @@
+import datetime
 import logging
 from typing import List, Optional, Any, Tuple, Union
 
-from credsweeper.common.constants import RECURSIVE_SCAN_LIMITATION
+from credsweeper.common.constants import RECURSIVE_SCAN_LIMITATION, RuleType
```
babenek, line 5:
```suggestion
from credsweeper.common.constants import RECURSIVE_SCAN_LIMITATION
```
TODO: apply without CI before merge

credsweeper/deep_scanner/html_scanner.py
```diff
@@ -19,7 +19,8 @@ def data_scan(
             depth: int,  #
             recursive_limit_size: int) -> List[Candidate]:
         """Tries to represent data as html text and scan as text lines"""
-        if data_provider.represent_as_html(depth, recursive_limit_size):
+        if data_provider.represent_as_html(depth, recursive_limit_size,
+                                           self.scanner.keywords_required_substrings_check):
```
csh519, line 23: Other changes are looking good.
But I couldn't understand why `represent_as_html()` should to filter the keywords with required sub-strings... :thinking: 
> `HTML representation : collect matched keywords fields only`

babenek: The representation combine header of table with each cell in the column (--depth > 2 option).
To reduce extra combinations i put the check. If  a header has no any substring - there is no sense do the combination

csh519 (approved):
Multiple credential detecting function and new rule `PII` are added.

LGTM :+1:

iuriimet (approved):

PR #413, opened 2023-08-29
Bugfix: extra memory consumption during html representation

babenek:
## Description

- Fix issue with huge memory consumption during parsing git://git.tizen.org/platform/framework/web/chromium-efl/third_party/catapult/third_party/vinn/third_party/parse5/benchmark/spec.html with --doc option
- refactored represent_as_html to use depth and size limit

## How has this been tested?

- [x] UnitTest
- [x] scan the repo with --doc option


xDizzix (dismissed):
Recursion was limited

credsweeper/file_handler/data_content_provider.py
```diff
@@ -171,9 +173,107 @@ def _check_multiline_cell(self, cell: Tag) -> Optional[str]:
             # the cell will be analysed as multiline text
             self.line_numbers.extend(line_numbers)
             self.lines.extend(stripped_lines)
-            return " ".join(stripped_lines)
+            # additionally all line feeds in HTML transform to spaces
+            self.line_numbers.append(line_numbers[0])
+            self.lines.append(" ".join(stripped_lines))
```
csh519, line 178: Could you explain more, why `self.lines` and `self.line_numbers` are appended here?
It seems those two variables already have been extended above.

babenek: the duplication represent text as it is rendered 
https://github.com/Samsung/CredSweeper/blob/09137d8081f06c5e33b0ce3df120b65e1828fbb3/tests/samples/test.html#L82
In case of long line the join will be skipped, but previous cases are kept

xDizzix (approved):
+

credsweeper/file_handler/data_content_provider.py
```diff
@@ -152,28 +154,127 @@ def represent_as_xml(self) -> bool:
...
+        for offset, line in enumerate(cell_lines):
             if stripped_line := line.strip():
-                line_numbers.append(cell.sourceline)
+                line_numbers.append(cell.sourceline + offset)
                 stripped_lines.append(stripped_line)
         if 0 == len(stripped_lines):
             return None
         elif 1 == len(stripped_lines):
-            return stripped_lines[0]
+            return line_numbers[0], stripped_lines[0]
```
csh519, line 173: It seems returning value `line_numbers[0]` not used anywhere...
Could you please check this again?

credsweeper/file_handler/data_content_provider.py
csh519: We communicated regarding this issue and decided to exclude this feature on this PR.

credsweeper/file_handler/data_content_provider.py
babenek: Yes, it is used 4 times. I've refactored the code to reduce complexity.

csh519: Ah.. `td_numbered_line[]`.. I see :)

csh519 (approved):
Thank you for fixing bug.

LGTM :+1:

iuriimet (approved):

PR #422, opened 2023-09-15
Extract all candidates from single line for one rule

babenek:
## Description

- Fix case when only first candidate is extracted from single line for one rule
- slowdown ~30% due continious scan in single line

## How has this been tested?

- [x] UnitTest
- [x] todo: rollback custom benchmark after https://github.com/Samsung/CredData/pull/46
- [x] Benchmark


.github/workflows/benchmark.yml
```diff
@@ -62,7 +62,8 @@ jobs:
     - name: Checkout CredData
       uses: actions/checkout@v3
       with:
-        repository: Samsung/CredData
+        repository: babenek/CredData
+        ref: polycand
```
```suggestion
        repository: Samsung/CredData
```

credsweeper/scanner/scan_type/scan_type.py
```diff
@@ -89,43 +90,54 @@ def get_line_data(
...
+                for pattern in rule.patterns[1:]:
+                    aux_line_data_list = cls.get_line_data_list(config=config,
+                                                                target=target,
+                                                                pattern=pattern,
+                                                                filters=rule.filters)
+                    for aux_line_data in aux_line_data_list:
+                        if len(config.exclude_values) > 0 and aux_line_data.value.strip() in config.exclude_values:
+                            continue
+                        candidate.line_data_list.append(aux_line_data)
+                        break
```
csh519, line 136: Why this `for` loop just add only one `aux_line_data`?

credsweeper/scanner/scan_type/scan_type.py
```diff
@@ -89,43 +90,54 @@ def get_line_data(
...
+                    else:
+                        break
+                else:
+                    candidates.append(candidate)
+            else:
+                candidates.append(candidate)
```
csh519, line 142: These `if-else` and `for-else` chains are very hard to read...
Could you please refactor above code block by extracting methods?

credsweeper/scanner/scan_type/scan_type.py
babenek: You are right. Now there are appended at least two line data (left and/or right sides nearest from primary value)

credsweeper/scanner/scan_type/scan_type.py
babenek: Just one method added. So many arguments are passed through all thread :(

credsweeper/scanner/scan_type/scan_type.py
```diff
@@ -89,43 +90,75 @@ def get_line_data(
...
+                    nearest_right = line_data
+
+            # append the values from both sides
+            if nearest_left:
+                candidate.line_data_list.append(nearest_left)
+            if nearest_right:
+                candidate.line_data_list.append(nearest_right)
+            # if none of both values - break the auxiliary scan with failure
+            if not nearest_left and not nearest_right:
+                return False
```
csh519, line 161: Could you please explain why does `_aux_scan()` is needed?

I tried to understand the code but i can't find out the connection between `line_data`s with `candidate.line_data_list[0]`..
Why it needed to find `nearest_left` and `nearest_right`?

babenek: It reduces possible possible wrong secondary values in a line. Only two values might be taken for primary value. From left and from right.

csh519 (approved):
Thank you for improving the scanning logic.

LGTM :+1:

csh519:
BTW this PR can be affected to other service..
Benchmark testing time should be improved.

Yullia (approved):

PR #426, opened 2023-09-25
Skip XMLParsedAsHTMLWarning from beautifulsoup

babenek:
## Description

- Suppress extra warning during parsing XML with --doc option:
```
bs4/builder/__init__.py:545: XMLParsedAsHTMLWarning: It looks like you're parsing an XML document using an HTML parser. If this really is an HTML document (maybe it's XHTML?), you can ignore or filter this warning. If it's XML, you should know that using an XML parser will be more reliable. To parse this document as XML, make sure you have the lxml package installed, and pass the keyword argument `features="xml"` into the BeautifulSoup constructor.
  warnings.warn(
```

## How has this been tested?

- [x] ``python -m credsweeper --doc --path tests/samples/ --save-json /dev/null``


csh519 (approved):
`XMLParsedAsHTMLWarning` is not useful warning, it should be ignored for clean logging.

LGTM :+1:

Yullia (approved):

PR #438, opened 2023-10-14
Key storage detecting

babenek:
## Description

based on #441

- Add JKS and PKCS12 storages detection in --doc mode
- 3 default passwords are used: empty, "changeit" and "changeme" to open a storage

## How has this been tested?

- [x] UnitTest
- [x] Benchmark - does not impact
- [x] fuzzing


csh519 (approved):
Thank you for adding `jks` and `pkcs12` scanner.

LGTM :+1:

credsweeper/deep_scanner/jks_scanner.py
```diff
@@ -0,0 +1,38 @@
...
+            self,  #
+            data_provider: DataContentProvider,  #
+            depth: int,  #
+            recursive_limit_size: int) -> List[Candidate]:
+        """Tries to scan JKS to open with standard password"""
+        candidates = []
+        for pw_probe in ["", "changeit", "changeme"]:
+            try:
+                keystore = jks.KeyStore.loads(data_provider.data, pw_probe, try_decrypt_keys=True)
+                if keystore.private_keys or keystore.secret_keys:
```
kmnls, line 27: This part is not clear.
You check for private keys in the keystore. They are returned in encrypted form (if your password guess is wrong) or decrypted format (if password matches). This is the first part of IF.
Then ELSE section is used for cases when there are no private keys at all. But you marked jks as candidate as well.
It seems there is a trouble out here. 4 cases should be checked independently:
- pkey present, default password, issue.
- pkey present, unknown password, no issue.
- no pkey, default password, no issue (?)
- no pkey, unknown password, no issue.

babenek: In any case - a storage with default password should not appear. 
IF-ELSE block just decide which text will be in info for user.

Unknown password will raise an exception and no issue is created.

tests/samples/dummy.jks
kmnls, line 1: Whether it is used? Cannot see traces in the code.

kmnls (commented):
See comment.
The code does not match jks source code behavior https://pyjks.readthedocs.io/en/latest/_modules/jks/jks.html#KeyStore.loads

credsweeper/deep_scanner/jks_scanner.py
kmnls: https://pyjks.readthedocs.io/en/latest/_modules/jks/jks.html#KeyStore.loads
                if try_decrypt_keys:
                    try:
                        entry.decrypt(store_password)
                    except DecryptionFailureException:
                        pass # ok, let user call decrypt() manually
                        
                        
No exception

babenek: In common (wrong) case the password of keys is the same like for the storage.
Suppose, if the key has good encryption - else branch will be taken.
Anyway, the scanner finds unencrypted(weak password) storage.

xDizzix (approved):
+

PR #441, opened 2023-10-16
Slight changes for development

babenek:
## Description

- Optimize ML validator creation - might help in multijob usage, e.g. #424
- Reorganize parent classes in DeepScanner to add new scanners in separated lines and reduce possible conflicts
- optimize file signature detection: startswith instead each element comparison

## How has this been tested?

- [x] UnitTest
- [x] Benchmark - no benchmark impact

### DIFF
--- a/credsweeper/app.py
+++ b/credsweeper/app.py
@@ -158,12 +158,13 @@ class CredSweeper:
             logger.info("ML validation is disabled")
             return False
         if not self.credential_manager.candidates:
-            logger.info("Skipping ML validation due to no candidates found")
+            logger.info("Skip ML validation because no candidates were found")
             return False
         for i in self.credential_manager.candidates:
             if i.use_ml:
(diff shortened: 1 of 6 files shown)


Yullia (dismissed):

xDizzix (approved):
Good

babenek:
Rollback ML optimization for #424 only

csh519 (approved):
New code block of `_use_ml_validation()` seems duplicated with `post_processing()`'s one.
But i assume it will be used on next PR.

LGTM :+1:

babenek:
> New code block of `_use_ml_validation()` seems duplicated with `post_processing()`'s one. But i assume it will be used on next PR.
> 
> LGTM 👍

Partially, it is.
There is possibly case when the same value is available for ml but after grouping it will be skipped.

true13:
Good

PR #443, opened 2023-10-19
Update Readme and docs

vpolyukh:
## Description

Update the response example of the found credential


babenek (approved):
Thank you for ocumentation updating!

csh519 (approved):
Thank you for update the docs.

LGTM :+1:

PR #449, opened 2023-11-01
PASS keyword false positive fix

babenek:
## Description

- Add fix for PASS keyword from many FP cases

## How has this been tested?

- [x] UnitTest
- [x] Benchmark


csh519:
Does it only for reducing 1 FP case?
Or did you intended to reduce FPs of internal cases?

babenek:
@csh519 
Yes, the fix is proposed after internal FP case. One benchmark FP was likely removed too.

csh519 (approved):
Thank you for reduce FP of `Password` rule.

LGTM :+1:

iuriimet (approved):

PR #456, opened 2023-11-18
PEM private keys detect enchancement

babenek:
## Description

- "filter_type" is an optional field in rules.
- PEM private key Elliptic Curves are found also
- strict filtering for PEM keys (ASN1 or keyword "bcrypt")
- Encrypted keys are skipped
- PGP keys are checked with entropy only

## How has this been tested?

- [x] UnitTest
- [x] Benchmark
- [x] TODO: rollback custom benchamark after https://github.com/Samsung/CredData/pull/52


credsweeper/rules/config.yaml
```diff
@@ -386,9 +386,7 @@
   severity: high
   type: pem_key
   values:
-    - (?P<value>-----BEGIN\s(?!ENCRYPTED|EC)[^-]*PRIVATE[^-]*KEY[^-]*-----(.+-----END[^-]+-----)?)
-  filter_type:
-    - LineSpecificKeyCheck
+    - (?P<value>-----BEGIN\s(?!ENCRYPTED)[^-]*PRIVATE[^-]*KEY[^-]*-----(.+-----END[^-]+-----)?)
```
csh519, line 389: Could you explain it more about why you decided to remove `EC` keyword from the pattern?
I think EC(elliptic curve encryption) can't be decrypted..

babenek: In openssl/test exists not encrypted EC key https://github.com/openssl/openssl/blob/master/test/testec-p256.pem.
It can be analysed with ASN1 parser. It works without a password:
```
openssl dgst -sha256 -sign testec-p256.pem -out signature.bin smcont.txt 
openssl dgst -sha256 -verify testecpub-p256.pem  -signature signature.bin smcont.txt 
```
Encrypted key i found has structure like this and cannot be parsed with asn1:
```
-----BEGIN EC PRIVATE KEY-----
Proc-Type: 4,ENCRYPTED
DEK-Info: AES-128-CBC,692.....

BXv....
```
So, i think, with the simple analysis for ASN1 structure we can decide whether a key is encrypted or not in the level.

csh519: Oh.. i didn't know about that.
Yes, using ASN1 decryption can be a solution.
How about change the logic to check ASN1 decryptable if the line includes EC keyword?
And other cases that doesn't include EC keyword just follow the logic before.

credsweeper/rules/config.yaml
```diff
@@ -326,12 +326,10 @@
   min_line_len: 72
 
 - name: PEM Private Key
-  severity: high
+  severity: critical
```
babenek, line 329:
```suggestion
  severity: high
```

Yullia (approved):

kmnls (dismissed):
Agree

.github/workflows/benchmark.yml
```diff
@@ -62,7 +63,8 @@ jobs:
     - name: Checkout CredData
       uses: actions/checkout@v3
       with:
-        repository: Samsung/CredData
+        ref: ecpkey
+        repository: babenek/CredData
```
babenek, line 67:
```suggestion
        repository: Samsung/CredData
```

babenek (commented):
TODO: rollback benchmark reference before merge

babenek:
@csh519 @kmnls @Yullia @xDizzix @iuriimet 
please approve https://github.com/Samsung/CredData/pull/52 first

Yullia (approved):

iuriimet (approved):

PR #465, opened 2023-11-26
Delete .github/workflows/benchmark.yml

babenek:
## Description

- Removed benchmark step to avoid any data leak

## How has this been tested?

- [x] UnitTest


csh519 (changes requested):
I think checking the benchmark score on PR is necessary.
Please don't merge it before we come up to conclusion.

PR #468, opened 2023-11-29
Use value offset in report sorting

babenek:
## Description

- Use value_start in sorting to prevent variouos order in tests
- Round ml_probability to 5 digits in test and the sort like previous applied

## How has this been tested?

- [x] UnitTest - updated jsons with result
- [x] Benchmark - no impact


deadkittten (approved):
"value" field changed to the start-end diapason, tests was updated accordingly to this change.
Looks good.

kmnls (approved):
Agree

PR #470, opened 2023-12-03
Separate function for future reusing

babenek:
## Description

- base64 decode function is extracted for reuse
- applied padding auto augmentation for base64 JWT which may be without padding
- ASN1 size check applied for small size case

## How has this been tested?

- [x] UnitTest
- [x] Benchmark


Yullia (approved):

credsweeper/utils/util.py
```diff
@@ -613,3 +616,17 @@ def parse_python(source: str) -> List[Any]:
...
+
+    @staticmethod
+    def decode_base64(text: str, padding_safe: bool = False, urlsafe_detect=False) -> bytes:
+        """decode text to bytes with / without padding detect and urlsafe symbols"""
+        value = text
+        if padding_safe:
+            pad_num = 0x3 & len(value)
+            if pad_num:
+                value += '=' * (4 - pad_num)
+        if urlsafe_detect and '-' in value or '_' in value:
```
kmnls, line 628:
```suggestion
        if urlsafe_detect and ('-' in value or '_' in value):
```
AND has higher precedence then OR

kmnls (changes requested):
See comment

babenek:
> See comment

Thank you for review. The issue was fixed.

xDizzix (approved):
+

kmnls (approved):
Agree

PR #478, opened 2023-12-14
Base64 encoded values in plain text detection

babenek:
## Description

- Add BASE64 PEM private key pattern with filter to detect encoded private key
- Separated PEM KEY detection to Util to avoid circular dependency issue

## How has this been tested?

- [x] UnitTest
- [x] Benchmark
- [x] TODO: rollback after https://github.com/Samsung/CredData/pull/61 merge


.github/workflows/benchmark.yml
```diff
@@ -62,7 +62,8 @@ jobs:
     - name: Checkout CredData
       uses: actions/checkout@v3
       with:
-        repository: Samsung/CredData
+        ref: LS0t
+        repository: babenek/CredData
```
```suggestion
        repository: Samsung/CredData
```

Yullia (approved):

kmnls (approved):
Agree

PR #482, opened 2024-01-02
Convert script from old tensorflow model to onnx

babenek:
## Description

- Add script for model convertion
- Converted the model again with modern packages
- Add integrity model check
- Remove git workflow check because squash merge is used

## How has this been tested?

- [x] UnitTest
- [x] Benchmark

https://netron.app/

| WAS | BECAME |
| --- | --- |
| ml_model onnx-old | ml_model onnx-new |


kmnls (approved):
Agree

meanrin (approved):
Seems ok, only major change is that bidirectional LSTM is now a single instance instead of two separate instances. Which is a good change ~
Thanks for adding the conversion script as well

PR #505, opened 2024-02-07
Fix url credential and openssh false positives

babenek:
## Description

- Fix url credentials FP
- Fix empty OPENSSH keys FP

## How has this been tested?

- [x] UnitTest
- [x] Benchmark


kmnls (commented):
Why do you obfuscate some connection strings?
To avoid some checker?
Even if they false-positive, they still must be detected anyway.

babenek:
> Why do you obfuscate some connection strings? To avoid some checker? Even if they false-positive, they still must be detected anyway.
%6D - not an obfuscation. Just to check % sign is caught

csh519 (approved):
Thank you for fix the FPs.

LGTM :+1:

iuriimet (approved):

PR #508, opened 2024-02-09
Change model with autokeras lstm v2

Yullia:
## Description

ML model updated with updated markup and parameters.
| WAS | BECAME |
| --- | --- |
| ml_model_old_arch | ml_model_arch |


babenek:
The approach was copied in #610. Thank you for the research

PR #522, opened 2024-03-03
Scan oversized lines by chunks

babenek:
## Description

- Add chunk-based scan for oversized lines with overlap to do not skip a credential between chunks
- slowdown ~26% (in the full dataset)
- slowdown ~ 2.2% for dataset without ovesized lines

## How has this been tested?

- [x] UnitTest
- [x] Benchmark


.github/workflows/benchmark.yml
```diff
@@ -22,7 +22,12 @@ jobs:
       - name: Checkout CredData
         uses: actions/checkout@v4
         with:
-          repository: Samsung/CredData
+          repository: babenek/CredData
+          ref: tachunks
+
+      - name: Markup hashing
+        run: |
+          for f in $(find meta -type f|sort); do md5sum $f; done >meta.md5
```
babenek, line 30:
```suggestion
```

.github/workflows/benchmark.yml
```diff
@@ -66,7 +71,12 @@ jobs:
...
```
babenek, line 79:
```suggestion
```

.github/workflows/benchmark.yml
```diff
@@ -171,7 +181,12 @@ jobs:
...
```
babenek, line 189:
```suggestion
```

.github/workflows/benchmark.yml
```diff
@@ -348,7 +363,12 @@ jobs:
       - name: Checkout CredData
         uses: actions/checkout@v3
         with:
-          repository: Samsung/CredData
+          repository: babenek/CredData
+          ref: tachunks
+
+      - name: Markup hashing
+        run: |
+          for f in $(find meta -type f|sort); do md5sum $f; done >meta.md5
```
babenek, line 371:
```suggestion
```

babenek (commented):
rollback

csh519 (commented):
Thank you for suggest great idea to scanning oversize lines!
It seems there is a problem when scanning full benchmark data...
So how about set oversize limit const like as below?

``` python
OVERSIZE_LINE_LIMIT = MAX_LINE_LENGTH * 10  # You can change the number.
```

I think it would be good to find a number which is 10 above, by testing with benchmark data. I hope the slowdown should under 10%..

babenek:
todo: rollback custom BM ref after https://github.com/Samsung/CredData/pull/142

babenek:
@csh519 
The overlap duplicates issue is fixed.
https://github.com/Samsung/CredSweeper/pull/522/files#diff-a308e37dd0360462596f1887de9d165bf5fb16132a748481cc54a1b64b77b62f
- two the same Candidates were used in the test. The purging is processed with unique positions of value, variable, separator and line.

csh519 (dismissed):
Oversize lines can be scanned after this PR.
Thank you for suggestion!

LGTM :+1:

xDizzix (approved):

iuriimet (approved):

PR #574, opened 2024-06-26
Discord bot token and Grafana service

babenek:
## Description

- Add Discord bot token pattern and filter ([test sample](https://github.com/discord/sample-game-integration/blob/master/server/example.cfg))
- Add Grafana service token pattern and filter (tested for tokens from [opensource repos](https://github.com/grafana/grafana/blob/main/public/openapi3.json))
- skip .map  .rc   .vdproj   .rc2   files from scan
- actualize Facebook token pattern
- fix quotation keyword pattern (skip some filters for well quoted values)

## How has this been tested?

- [x] UnitTest
- [x] Benchmark

### DIFF
--- a/credsweeper/rules/config.yaml
+++ b/credsweeper/rules/config.yaml
@@ -1244,7 +1244,7 @@
   confidence: strong
   type: pattern
   values:
-    - (?<![0-9A-Za-z_/+-])(?P<value>[NMO][ADgjQTwz][a-zA-Z0-9_-]{22,26})\.[a-zA-Z0-9_-]{6}\.[a-zA-Z0-9_-]{30,40}(?![0-9A-Za-z_/+-])
+    - (?<![0-9A-Za-z_/+-])(?P<value>[NMO][ADgjQTwz][a-zA-Z0-9_-]{22,26}\.[a-zA-Z0-9_-]{6}\.[a-zA-Z0-9_-]{30,40})(?![0-9A-Za-z_/+-])
   min_line_len: 62
   filter_type:
     - ValueDiscordBotCheck
--- a/tests/data/depth_3.json
+++ b/tests/data/depth_3.json
@@ -1140,16 +1140,16 @@
                 "line_num": 1,
                 "path": "tests/samples/discord_bot_token",
                 "info": "tests/samples/discord_bot_token|RAW",
-                "value": "MTIzNDU2Nzg5MDEyMzQ1Njc4OQ",
+                "value": "MTIzNDU2Nzg5MDEyMzQ1Njc4OQ.E2-E4_.Zig9V5mpMk-JybgCFvqSfgY9EoqWjkA5O_qDje",
                 "value_start": 0,
-                "value_end": 26,
+                "value_end": 72,
                 "variable": null,
                 "variable_start": -2,
                 "variable_end": -2,
                 "entropy_validation": {
                     "iterator": "BASE64_CHARS",
-                    "entropy": 3.997952698660691,
-                    "valid": false
+                    "entropy": 4.731746181697384,
+                    "valid": true
                 }
             }
         ]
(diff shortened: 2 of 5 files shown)


xDizzix (approved):
+

credsweeper/common/constants.py
```diff
@@ -17,8 +17,10 @@ class KeywordPattern:
     # Authentication scheme ( oauth | basic | bearer | apikey ) precedes to credential
     value = r"(?P<value_leftquote>((b|r|br|rb|u|f|rf|fr|\\{0,8})?[`'\"]){1,4})?" \
             r"( ?(oauth|bot|basic|bearer|apikey|accesskey) )?" \
-            r"(?P<value>(?:\{[^}]{3,8000}\})|(?:<[^>]{3,8000}>)|" \
-            r"(?(value_leftquote)(?:\\[tnrux0-7][0-9a-f]*|[^`'\"\\])|(?:\\n|\\r|\\?[^\s`'\"\\,;])){3,8000})" \
+            r"(?P<value>" \
+            r"(?(value_leftquote)(?:\\[tnrux0-7][0-9a-f]*|[^`'\"\\])|(?:\\n|\\r|\\?[^\s`'\"\\,;])){3,8000}" \
+            r"|(?:\{[^}]{3,8000}\})|(?:<[^>]{3,8000}>)" \
+            r")" \
```
csh519, line 23: Could you explain why those lines are switched?

credsweeper/rules/config.yaml
```diff
@@ -644,7 +659,7 @@
   confidence: moderate
   type: pattern
   values:
-    - (?<![0-9A-Za-z_+-])(?P<value>SK[0-9a-fA-F]{32})(?![=0-9A-Za-z_+-])
+    - (?<![0-9A-Za-z_+-])(?P<value>SK[0-9A-Fa-f]{32})(?![=0-9A-Za-z_+-])
```
csh519, line 662: Do they have difference? Did you changed it just for beautify?

.github/workflows/benchmark.yml
```diff
@@ -160,14 +160,14 @@ jobs:
...
-    # put the benchmark in single job to keep constant environment during test
+    # put the benchmark in single job to keep constant environment during test python 3.8 is not applicable
     needs: [ download_data ]
 
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
-        python-version: [ "3.10", "3.9", "3.10", "3.11" ]
+        python-version: [ "3.9", "3.10", "3.11" ]
```
csh519, line 170: If the benchmark is not available on python 3.8, then should we consider to set minimum version of CredSweeper as 3.9?

babenek: There only speed benchmark limitation. CredSweeper can work wit 3.8.
As i see, https://pypistats.org/packages/credsweeper - the version is most popular. 
With any limitation of a package we will set the minimum to 3.9

csh519: Okay. 3.8 is the most popular version...

credsweeper/common/constants.py
babenek: that regex collects full quoted value first "``[2ns8BH]ds$%^``"
previous - collects only in brackets "``[2ns8BH]``ds$%^"

credsweeper/rules/config.yaml
babenek: Yes, i try to unify the pattern with all rules to quick search them in future
The order relates ASCII table.

csh519: Good idea.

credsweeper/common/constants.py
csh519: Hmm... why do you think bracket included value is correct value?
I think previous regex was intended to filter the bracket included values.

kmnls (approved):
Agree

Yullia (approved):
Approved

credsweeper/common/constants.py
babenek: https://github.com/Samsung/CredSweeper/actions/runs/9839488594/job/27161670040
- the test with random generation

csh519: Okay i see.

PR #595, opened 2024-08-11
ValuePathFilter and keyword regex enchancement

babenek:
## Description

- Add square brackets in keyword regex to detect a list definition
- Add case when keyword is in protocol name and followed URL is obtained like a variable with semicolon as delimiter ``authp://some.url``
- case of TOML definitions

## How has this been tested?

- [x] UnitTest
- [x] Benchmark todo: after https://github.com/Samsung/CredData/pull/159


xDizzix (approved):
+

Yullia (approved):
Great job!

PR #600, opened 2024-08-27
New patterns & ML update

babenek:
## Description

- Refactoring ML features with reducing DC and added regex matching for an attribute
- Retrain ML. summary F1 = 0.952886
- Added new ML features for using regex in an attribute
- ML training produces chart with MD5 sums of config and model in bottom to manage them
- Removed VariableNotAllowedPatternCheck filter but apply the pattern in ML. Variuos case with _ID postfixes and AWS keys.
- Refactored Keyword pattern to deal with escaped quotes in a value. Tests added. Some cases still are not covered.
- Filter ValuePath was extended with morphemes check to reduce FN
- Removed duplicated filters

## How has this been tested?

- [x] UnitTest
- [x] Benchmark


.github/workflows/benchmark.yml
```diff
@@ -22,7 +22,8 @@ jobs:
       - name: Checkout CredData
         uses: actions/checkout@v4
         with:
-          repository: Samsung/CredData
+          repository: babenek/CredData
```
babenek, line 25: TODO: rollback after https://github.com/Samsung/CredData/pull/164

babenek:
@Samsung/credsweeper_maintainers , please give your opinion about the rules names.

.github/workflows/benchmark.yml
```diff
@@ -72,7 +73,8 @@ jobs:
...
+          ref: auxiliary
```
babenek, line 77:
```suggestion
          repository: Samsung/CredData
```

.github/workflows/benchmark.yml
```diff
@@ -169,7 +171,8 @@ jobs:
...
```
babenek, line 175:
```suggestion
          repository: Samsung/CredData
```

.github/workflows/benchmark.yml
```diff
@@ -351,7 +354,8 @@ jobs:
...
```
babenek, line 358:
```suggestion
          repository: Samsung/CredData
```

babenek (commented):
TODO: rollback after https://github.com/Samsung/CredData/pull/167

babenek:
bencmark fails before https://github.com/Samsung/CredData/pull/167

credsweeper/credentials/line_data.py
```diff
@@ -191,9 +191,17 @@ def clean_bash_parameters(self) -> None:
                 self.value = value_whsp[0]
 
     def clean_toml_parameters(self) -> None:
-        """Curly brackets may be caught in TOML format"""
-        while self.value.endswith('}') and '{' in self.line[:self.value_start]:
-            self.value = self.value[:-1]
+        """Parenthesis, curly and squared brackets may be caught in TOML format and bash. Simple clearing"""
+        dirty = self.value and self.value[-1] in ['}', ']', ')']
```
csh519, line 195: `dirty` is also good name but IMO it would be good to re-name with like as `is_OOO`.
How about `is_cleared` or `is_clear`?

credsweeper/filters/value_method_check.py
```diff
@@ -30,6 +30,6 @@ def run(self, line_data: LineData, target: AnalysisTarget) -> bool:
         """
         if line_data.is_well_quoted_value:
             return False
-        if "function" in line_data.value or self.PATTERN.search(line_data.value):
+        if "function" in line_data.value or self.PATTERN.match(line_data.value):
```
csh519, line 33: Is there any FP cases that include `function` keyword in the value?
I think checking `function` keyword by `search()` seems reasonable..

credsweeper/filters/variable_not_allowed_pattern_check.py
```diff
@@ -1,36 +0,0 @@
-import re
```
csh519, line 1: Why this filter has been removed?
There is no description for it..

csh519 (commented):
Please check my comments below.

And next time, please separate PR by feature and request review.
As you know huge PR couldn't have good review.

credsweeper/filters/value_method_check.py
babenek: actually, there was FN case in random generated credentials like "12#@5(!)fs", so I refactored the pattern and applied ``match`` to use the search from begin of a value

credsweeper/filters/variable_not_allowed_pattern_check.py
babenek: AWS_KEY_ID is filtered often :(
The feature was migrated to ML

babenek: VariableNotAllowedPatternCheck mention was added to PR description

credsweeper/filters/value_method_check.py
csh519: Hmm.. still i can get it clearly.
Do you mean there is a random generated credential value which includes "function" keyword in it?
Like as "12#@5(!)function3w@91"?
If then i think `search()` method working good.

babenek: No, i mean the second part of OR condition in the check. 
The ``function`` check does not use regex.

csh519: Ah.. I just misunderstood about this. 
But in some case like below, `search()` method still more useful doesn't it?

``` python
myPassword = "this_is_not_password"
```

credsweeper/filters/variable_not_allowed_pattern_check.py
csh519: Okay I see..
Is there a lot of patterns that do not allowed?
I think if we delegate this filter's role to ML, we will be harder to reasoning and debugging why.
So I'm worry about this.

credsweeper/filters/value_method_check.py
babenek: The filter does not return True for the line. Only if it will be 
```python
myPassword = this_is_not_password()
```
or
```bash
myPasswsord=password_function
```
*depends on file

Otherside, your code line means that a string is assigned for variable with name "password" 
So, it may be a password. Even "1234"  may be a password, but CredSweeper skips the sequence.

csh519: Okay i agree. Let's keep the change then.

credsweeper/filters/variable_not_allowed_pattern_check.py
csh519: It seems filtering the credentials from ML will be better..
Okay let's move the role to ML.

xDizzix (approved):

Yullia (approved):
Approved

xDizzix (approved):

PR #625, opened 2024-12-02
Hashicorp Vault Token

babenek:
## Description

- Add Hashicorp Vault Token pattern https://www.hashicorp.com/blog/vault-1-10 (expired time  is 24h max)

## How has this been tested?

- [x] UnitTest
- [x] Benchmark


kmnls (approved):
Agree

PR #643, opened 2024-12-19
Avoid install fail

babenek:
## Description

- Fix dbg which fails

## How has this been tested?

- [ ] release workflow


babenek:
force push for build fix

PR #644, opened 2024-12-22
fix xlsx report for diff

babenek:
## Description

- Add deprecation for main scan function
- Fix Bug for xlsx of diff
- Modify report names of json for diff (py3.8 pathlib does not support stem manipulation)
- Update tests for diff report verification
- Use prefix ``.added`` instead name ends ``_added`` for JSON diff reports

## How has this been tested?

- [x] UnitTest
- [x] Benchmark


credsweeper/__main__.py
```diff
@@ -270,20 +271,30 @@ def get_json_filenames(json_filename: str):
...
+         xlsx_filename: Optional[str] = None) -> int:
     """Scan content_provider data, print results or save them to json_filename is not None
 
     Args:
         args: arguments of the application
         content_provider: FilesProvider instance to scan data from
-        json_filename: json type report file path or None
-        xlsx_filename: xlsx type report file path or None
+        json_filename: (Deprecated) json type report file path or None
+        xlsx_filename: (Deprecated) xlsx type report file path or None
```
csh519, line 284: It seems `json_filename` and `xlsx_filename` can be removed from `scan()`..
Could you please consider to remove those two parameters?

credsweeper/__main__.py
```diff
@@ -270,20 +271,30 @@ def get_json_filenames(json_filename: str):
     return added_json_filename, deleted_json_filename
```
csh519, line 271: It seems there `get_json_filenames()` doesn't seem to be used anywhere..
Could you please remove this function?

csh519 (approved):
Thank you for fix the bug of xlsx file diff.

LGTM 👍

kmnls (approved):
Agree

PR #951, opened 2026-09-20
Fix CRX3 ZIP payload extraction

arunimshukla:
## Description

- Parse CRX2 and CRX3 headers using their version-specific layouts.
- Reject unsupported versions and truncated headers instead of returning a misaligned payload.
- Add regression coverage for valid CRX2 and CRX3 payloads plus malformed inputs.

Before this change, `CrxScanner.match` accepted CRX3 files, but `zip_extract` always interpreted bytes 8–15 as the CRX2 public-key and signature lengths. For CRX3, bytes 8–11 are one protobuf-header length followed by the header, so the scanner could calculate the wrong ZIP offset and miss credentials inside modern Chrome extensions.

Chromium format reference: https://chromium.googlesource.com/chromium/src/+/main/components/crx_file/crx3.proto

## How has this been tested?

- [x] UnitTest
- [x] Benchmark

```
python -m pytest tests/deep_scanner/test_crx_scanner.py -q
# 4 passed

python -m pytest tests/deep_scanner -q
# 96 passed

yapf --style .style.yapf --diff credsweeper/deep_scanner/crx_scanner.py tests/deep_scanner/test_crx_scanner.py
flake8 credsweeper/deep_scanner/crx_scanner.py tests/deep_scanner/test_crx_scanner.py
pylint --py-version=3.12 credsweeper/deep_scanner/crx_scanner.py
# 10.00/10

mypy --config-file .mypy.ini --python-version=3.12 credsweeper
# Success: no issues found in 175 source files
```

Benchmark was not run because this changes header parsing only and does not alter detection rules or model scoring.


babenek:
Hi @arunimshukla,
thanks you for the fix! It really works with various CRX files and the code is ok.
Unfortunately, not all action tests are passed. I prepared the fix https://github.com/arunimshukla/CredSweeper/pull/1 - apply it in web application to make a last signed commit in the PR or merge it with fast-forward.
Then the actions may be started.
Best Regards

arunimshukla:
Hi @babenek, thanks for preparing the fix and for testing it against different CRX files.

I've merged your updates from arunimshukla#1 into this branch (2cf200c), so it's now in sync with the latest `main`. It's ready for the Actions to run whenever you are. If anything else fails, let me know and I'll address it.

babenek (dismissed):
Thank you for the fix!

credsweeper/deep_scanner/crx_scanner.py
```diff
@@ -22,10 +22,19 @@ def match(data: bytes | bytearray) -> bool:
...
+        if 2 == version:
+            pubkey_length = struct.unpack_from("<I", data, offset=8)[0]
+            signature_length = struct.unpack_from("<I", data, offset=12)[0]
+            zip_offset = 16 + pubkey_length + signature_length
+        elif 3 == version:
+            header_length = struct.unpack_from("<I", data, offset=8)[0]
+            zip_offset = 12 + header_length
+        else:
+            raise ValueError(f"Unsupported CRX version: {version}")
+        if len(data) < zip_offset:
```
babenek:
```suggestion
        if len(data) <= zip_offset:
```

.github/workflows/check.yml
```diff
@@ -96,7 +96,7 @@ jobs:
       - name: Banner and version check
         if: ${{ always() && steps.setup_credsweeper.conclusion == 'success' }}
         env:
-          CREDSWEEPER_BANNER: "CredSweeper 1.18.4 crc32:10ca2071"
+          CREDSWEEPER_BANNER: "CredSweeper 1.18.4 crc32:7a7fdaf7"
```
babenek:
```suggestion
          CREDSWEEPER_BANNER: "CredSweeper 1.18.4 crc32:646507f3"
```

credsweeper/deep_scanner/crx_scanner.py
```diff
@@ -22,10 +22,19 @@ def match(data: bytes | bytearray) -> bool:
...
+            raise ValueError(f"CRX header exceeds file size: {zip_offset} > {len(data)}")
```
babenek:
```suggestion
        if len(data) <= zip_offset:
            raise ValueError(f"CRX header offset ({zip_offset}) exceeds file size ({len(data)})")
```

babenek (approved):

Pomidorry (approved):
+

xDizzix (approved):

